PR #65, opened 2014-12-04
Adaptive concurrency

SzymonPobiega:
This pull changes the way SqlServer transport does polling so that the number of threads is dynamically adapted to the load


pawelpabich:
Done already. Great! When are you planning to release it to nuget ?

SzymonPobiega:
Good question. Hope this week

PR #83, opened 2015-06-29
Hotfix 1.2.3

mauroservienti:
Backporting `custom schema` feature to SQL Transport v1.x.
Related to: https://github.com/Particular/FeatureDevelopment/issues/320

@Particular/engineers please review


mauroservienti:
The build is failing because the GitVersion targets are not where they should be. They are where they should be in the repo. What should I do to fix it? Thanks.

mauroservienti:
merging to kick off the publish process.

Issue #91, opened 2015-09-17
missing NServiceBus.Core dependency on 1.2.3


timbussmann:
[Version 1.2.3](https://www.nuget.org/packages/NServiceBus.SqlServer/1.2.3) of the SqlServer package misses dependencies to NSB.Core which seem to be removed by accident while previous patch versions of the package still have the dependency.

ramonsmits:
+1 I just noticed it too on Nuget while downgrading.

timbussmann:
fixed with hotfix 1.2.4

PR #92, opened 2015-09-24
support suppressing native transactions

timbussmann:
related to https://github.com/Particular/NServiceBus/issues/2885 , message senders need a way to send a message without enlisting in native transactions.

also fixes https://github.com/Particular/NServiceBus.SqlServer/issues/91

### DIFF
--- a/src/NServiceBus.SqlServer/NServiceBus.SqlServer.csproj
+++ b/src/NServiceBus.SqlServer/NServiceBus.SqlServer.csproj
@@ -13,7 +13,8 @@
     <FileAlignment>512</FileAlignment>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(SolutionDir)NServiceBus.snk</AssemblyOriginatorKeyFile>
-    <NuGetPackageImportStamp>a65d899d</NuGetPackageImportStamp>
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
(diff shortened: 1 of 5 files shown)


timbussmann:
@SzymonPobiega review appreciated :)

src/NServiceBus.SqlServer/NServiceBus.SqlServer.csproj
```diff
@@ -42,12 +43,12 @@
...
     </Reference>
     <Reference Include="Microsoft.CSharp" />
-    <Reference Include="NServiceBus, Version=4.7.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
-      <HintPath>..\packages\NServiceBus.Interfaces.4.7.5\lib\net40\NServiceBus.dll</HintPath>
+    <Reference Include="NServiceBus, Version=4.4.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
+      <HintPath>..\packages\NServiceBus.Interfaces.4.4.8\lib\net40\NServiceBus.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="NServiceBus.Core, Version=4.7.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
-      <HintPath>..\packages\NServiceBus.4.7.5\lib\net40\NServiceBus.Core.dll</HintPath>
```
weralabaj: @timbussmann @SzymonPobiega  In this case are we downgrading NSB version? Could there be some subtle differences between 4.4 and 4.7 that can cause issues?

SzymonPobiega: Good catch .Downstream repos should reference the lowest minor of the major they are targeting in order to ensure they don't depend on features added in later minors. 

Also, they should reference the highest compatible version of acceptance tests.

src/NServiceBus.SqlServer/SqlServerMessageSender.cs
```diff
@@ -61,7 +64,9 @@ public void Send(TransportMessage message, Address address)
                     schemaName = SchemaNameCollection[ address.Queue ];
                 }
 
-                if (UnitOfWork.HasActiveTransaction(queueConnectionString))
+                bool suppressNativeTransactions;
+                PipelineExecutor.CurrentContext.TryGet("do-not-enlist-in-native-transaction", out suppressNativeTransactions);
```
SzymonPobiega: How is this solved in 5.x? Are we using `DoNotEnlistInReceiveTransaction`?

src/NServiceBus.SqlServer/NServiceBus.SqlServer.csproj
timbussmann: version 1.2.3 had missing dependencies on the nuget package, but targeted 4.7 (probably by accident since that was moved in when removing ripple as far as I saw that). 1.2.2 supported 4.4 therefore I decided to "downgrade" the dependencies to not "force" users using older 1.2 versions do upgrade to NSB 4.7.

src/NServiceBus.SqlServer/SqlServerMessageSender.cs
timbussmann: I think it's this flag on the SendOptions here: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Unicast/DeliveryOptions.cs#L26

SzymonPobiega: Yeah, but this one also prevents enlisting in a `TransactionScope` so the behavior is slightly different. Is it by purpose?

timbussmann: seems so. Whats the usecase of suppressing tx with dtc?

SzymonPobiega: This is for users who just want to shoot a message regardless what is the outcome of the handler (i.e. from the `catch` clause). I am pretty sure that enlisting in a `TransactionScope` would not hurt here anyway but just thinking out loud if we should make these more similar in order for future selves to understand it better.

timbussmann: the approach for future solutions would be to check if not using dtc and only in this case suppress enlistment but adding the whole check in such old versions would be to much effort I think. Goal is definitely to remove this flag again since we should have the capabilities to reach the same goal in v6 (not sure about v5).

src/NServiceBus.SqlServer/NServiceBus.SqlServer.csproj
weralabaj: @timbussmann I'd double check with a person that targeted 4.7 first to confirm that it wasn't intentional (I guess it wasn't but who knows). If that was indeed by accident then it makes sense to downgrade.

timbussmann: let's see if @mauroservienti remembers, he introduced that NuGet dependency on that support branch (https://github.com/Particular/NServiceBus.SqlServer/commit/35b2b383d37e4ec10287f8950f303cf687114ac1)

mauroservienti: no good reason I can recall, other than using NuGet when removing ripple and then picking up latest at that time. The only thing I'd check is if ServiceControl can run it.

timbussmann: could someone check that then? I'd be very happy if someone else could do that since I'm completely busy preparing all the hotfix releases.

mauroservienti: @mikeminutillo @gbiellem or @johnsimons should have an answer for you

SzymonPobiega:
Good to go!

PR #94, opened 2015-11-05
Port Sql Transport to NSB v6

tmasternak:
Initial version of sql v3

Connects to #100


weralabaj:
@andreasohlund Few remarks:

- Could we make `MessageProcessingAbortedException` public? In our view that should be a part of the contract and we really need that for transport implementations. Maybe the name should be adjusted to something more meaningful to indicate this is retry logic and shouldn't be handled by clients?
- According to [docs](http://docs.particular.net/nservicebus/sqlserver/multiple-databases#current-endpoint) for schema config file has higher priority than code, for connection string it's the other way around. Inconsistent? Should we change that?
- Would it make sense to wrap and expose settings used by transport as properties? E.g. that looks really bad and is fragile:

```
protected override void ConfigureForReceiving(TransportReceivingConfigurationContext context)
{
  var transactionOptions = new TransactionOptions
  {
    IsolationLevel = context.Settings.Get<IsolationLevel>("Transactions.IsolationLevel"),
    Timeout = context.Settings.Get<TimeSpan>("Transactions.DefaultTimeout")
   }; 
}
```

@andreasohlund @SzymonPobiega @mauroservienti @danielmarbach The first version could use your reviews! (note the implementation is not finished - e.g. we don't have multi-db support yet)

danielmarbach:
I think we should switch to `OperationCancelledException` this would
essentially allow us to support proper cancellation.

On Tue, Nov 10, 2015 at 4:19 PM, Weronika Łabaj <notifications@github.com>
wrote:

> @andreasohlund <https://github.com/andreasohlund> Few remarks:
>
>    - Could we make MessageProcessingAbortedException public? In our view
>    that should be a part of the contract and we really need that for transport
>    implementations. Maybe the name should be adjusted to something more
>    meaningful to indicate this is retry logic and shouldn't be handled by
>    clients?
>    - According to docs
>    <http://docs.particular.net/nservicebus/sqlserver/multiple-databases#current-endpoint>
>    for schema config file has higher priority than code, for connection string
>    it's the other way around. Inconsistent? Should we change that?
>    - Would it make sense to wrap and expose settings used by transport as
>    properties? E.g. that looks really bad and is fragile:
>
> protected override void ConfigureForReceiving(TransportReceivingConfigurationContext context)
> {
>   var transactionOptions = new TransactionOptions
>   {
>     IsolationLevel = context.Settings.Get<IsolationLevel>("Transactions.IsolationLevel"),
>     Timeout = context.Settings.Get<TimeSpan>("Transactions.DefaultTimeout")
>    };
> }
>
> @andreasohlund <https://github.com/andreasohlund> @SzymonPobiega
> <https://github.com/SzymonPobiega> @mauroservienti
> <https://github.com/mauroservienti> @danielmarbach
> <https://github.com/danielmarbach> The first version could use your
> reviews! (note the implementation is not finished - e.g. we don't have
> multi-db support yet)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/NServiceBus.SqlServer/pull/94#issuecomment-155449169>
> .
>



-- 
*#NSBcon 2015 • The "All About NServiceBus" conference is back *
*• Learn more <http://go.particular.net/l/27442/2015-08-21/5d4wzp> →*

SzymonPobiega:
@weralabaj a bit of history: the connection string config predates the agreement we made among the former engineering group about a year ago that, should we ever need to configure something, the config file has precedence over code. The rationale behind this was that an operations person should be able to override whatever has been configured via code.

That is history. The new consensus (as I understood it) is that every new functionality will only have code configuration and if users want to use whatever config file they like, they need to implement this by themselves. An example of such approach is the new routing code-only API.

andreasohlund:
>I think we should switch to `OperationCancelledException` this would

I like it, lets go with that?

danielmarbach:
@tmasternak @weralabaj I was unable to do the review inside GH UI because of too many files changes. See review PR

https://github.com/Particular/NServiceBus.SqlServer/pull/101

tmasternak:
@andreasohlund @danielmarbach I like the name of the exception :+1:. My only concert is that making it public there will be nothing which would prevent handler logic from throwing that. Are we fine with it?

andreasohlund:
Pretty sure @danielmarbach meant to use https://msdn.microsoft.com/en-us/library/system.operationcanceledexception%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396 instead

tmasternak:
>Pretty sure @danielmarbach meant to use https://msdn.microsoft.com/en-us/library/system.operationcanceledexception%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396 instead

@andreasohlund can you chime in on https://github.com/Particular/NServiceBus/pull/3078. It looks that this change might have more consequences than expected.

Issue #100, opened 2015-11-12
Async/Await support


danielmarbach:
Task force: @tmasternak @weralabaj @justabitofcode @mauroservienti
Make changes to Sql Transport that will make it possible to use it with NServiceBus v6

### Impact assessment  
Any user who uses sql transport and will want to migrate to v6.

### Plan of Attack

- Features
 - ~~[Transactions support](https://github.com/Particular/NServiceBus.SqlServer/issues/138)~~ (done)
 - [Multi-database support](https://github.com/Particular/NServiceBus.SqlServer/issues/139) 
 - ~~[Multi-schema support](https://github.com/Particular/NServiceBus.SqlServer/issues/140)~~ (done)
 - ~~[TTBR](https://github.com/Particular/NServiceBus.SqlServer/issues/141)~~ (done)
 - ~~[Auto-creation of db structure](https://github.com/Particular/NServiceBus.SqlServer/issues/142)~~ (done)
- Testing
 - ~~[Acceptance tests from core, SqlTransport v2 and new for v3](https://github.com/Particular/NServiceBus.SqlServer/issues/130)~~ (done)
 - ~~[Performance testing](https://github.com/Particular/NServiceBus.SqlServer/issues/125)~~ (out-of-scope)
 - ~~[Wirecompat test](https://github.com/Particular/NServiceBus.SqlServer/issues/131)~~ (done)
 - ~~[Callback receivers](https://github.com/Particular/NServiceBus.SqlServer/issues/132)~~ (done)
- Documentation
 - ~~[Smoke tests for doc content](https://github.com/Particular/NServiceBus.SqlServer/issues/133)~~
 - ~~[Upgrade guide](https://github.com/Particular/NServiceBus.SqlServer/issues/134) https://github.com/Particular/docs.particular.net/issues/1195~~
 - ~~[Multi-schema v3 to v1/v2 setup guide](https://github.com/Particular/NServiceBus.SqlServer/issues/135)~~ (not relevant anymore)
 - ~~[Connection pool guidance](https://github.com/Particular/NServiceBus.SqlServer/issues/136)~~
 - ~~[Update docs](https://github.com/Particular/NServiceBus.SqlServer/issues/137)~~ (done)
[...]

PR #108, opened 2015-12-02
Support for custom sql connection factory

tmasternak:
Part of #100 
Reason we need this: https://github.com/Particular/NServiceBus.SqlServer/pull/86
Waiting for: https://github.com/Particular/NServiceBus/pull/3158

### DIFF
--- a/src/NServiceBus.SqlServer/SqlServerTransport.cs
+++ b/src/NServiceBus.SqlServer/SqlServerTransport.cs
@@ -3,6 +3,7 @@ namespace NServiceBus
     using System;
     using System.Collections.Generic;
     using System.Data.SqlClient;
+    using System.Threading.Tasks;
     using System.Transactions;
     using NServiceBus.Performance.TimeToBeReceived;
     using NServiceBus.Settings;
@@ -38,7 +39,7 @@ namespace NServiceBus
 
         SqlConnectionFactory CreateConnectoinFactory(string connectionString, ReadOnlySettings settings)
         {
-            Func<string, SqlConnection> factoryOverride;
+            Func<string, Task<SqlConnection>> factoryOverride;
 
             if (settings.TryGet(SqlServerSettingsKeys.ConnectionFactoryOverride, out factoryOverride))
             {
--- a/src/NServiceBus.SqlServer/SqlServerTransport.cs
+++ b/src/NServiceBus.SqlServer/SqlServerTransport.cs
@@ -2,6 +2,7 @@ namespace NServiceBus
 {
     using System;
     using System.Collections.Generic;
+    using System.Data.SqlClient;
     using System.Transactions;
     using NServiceBus.Performance.TimeToBeReceived;
     using NServiceBus.Settings;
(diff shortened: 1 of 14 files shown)


SimonCropp:
if we are giving them a delegate we dont need to pass them the connection string from config. they can read that in the delegate if they want to

tmasternak:
How would they get it is such case?

SimonCropp:
they own the connection string, either in code or in xml. no reason they should give it to use for us to give it back to them

tmasternak:
That is true until we have multi-database scenario. In that case connection string passed to lambda will change. We do not have multi-database yet but that should happen sooner than later. In addition current signature is compliant with old configuration api.

SimonCropp:
show me two code samples of how someone would configure 

1. all in code
2. in code combined with XML

as for multi database support what is the discriminator to switch between dbs?

SimonCropp:
also if you are already making design decisions based on multidatabase show me some config scenarios on how that will work

tmasternak:
I don't know how the scenarios will look in detail. That being said mapping transport addresses to connection strings (for multidatabase setup) and creating connection from a connection string are two different responsibilities and I think this should be reflected in extension points.

SimonCropp:
so since we dont know what multidatabase will look like lets not try and design for it now.

so i am back to: if they are in control of creating a connection then there is no value in us passing them the connection string they have passed us. it is pointless

tmasternak:
I think that we don't know the details but we do know something which influences the design.

To the second point. If user uses xml configuration he need to read the connection string one more time in lambda. Passing is more convenient and in addition we keep the old api unchanged.

SimonCropp:
u r missing the point. if they are taking control of the connection creation it is entirely likely that they want to just have it hard coded, or perhaps read it from some other config like yaml or json.

and yes if they want to read it from xml they can write one line of code to do so. but we dont want to leak app.confix based logic into a code based api. we avoid doing this everywhere else in the core

tmasternak:
Did you have a look at https://github.com/Particular/NServiceBus.SqlServer/pull/86? To my knowledge that is the main reason we need this.

src/NServiceBus.SqlServer/SqlServerTransport.cs
```diff
@@ -33,15 +35,27 @@ SqlServerAddressProvider CreateAddressParser(ReadOnlySettings settings)
             return parser;
         }
 
+        SqlConnectionFactory CreateConnectoinFactory(string connectionString, ReadOnlySettings settings)
```
andreasohlund: Typo

SimonCropp:
a simple delegate will still support  #86

tmasternak:
I agree the delegate would be fine. However what I'm pointing out is that self managing of connection string is not the reason this method was introduced.

SimonCropp:
> However what I'm pointing out is that self managing of connection string is not the reason this method was introduced.

ok but a simple delegate still allows us to support 86 while not taking a contradictory stance to how other apis work.  It also means less code overall

andreasohlund:
I'm with @SimonCropp here. I don't see the value of passing that string in, as Simon mentioned they can just read that them selfves if needed?

I'd say call YAGNI on this for now and handle this when we have more details on what we want to do with the multi data base support. (separate pull)

tmasternak:
I've removed connection string from factory.

PR #111, opened 2015-12-03
[WIP] Restructuring files to subfolders

tmasternak:


tmasternak:
Let's check if we should support Guards e.g. for configuration checks

Issue #110, opened 2015-12-03
Backwards compatibility testing scenarios


weralabaj:
### Compatibility Tests:
- [x]  Scenarios
 - [x] Simple send
 - [x] Response/Reply
 - [x] Pub/Sub
- [x] Configuration variations
 - [x] Versions 1, 2 and 3
 - [x] Single, multischema ~~and multidatabase~~
 - [x] Scale-out
- [x] Callbacks
 - [x] Simple
 - [x] Scale-out
- [x] Incompatible v2 config
- [x] Persistence. Sharing connections/transactions
 - Sagas
- [x] 3rd party integration - verify nothing changed
- [x] ServiceControl https://github.com/Particular/ServiceControl/issues/609

petersgiles:
@weralabaj how can I help?

tmasternak:
@weralabaj  @mauroservienti  I propose to remove the performance tests from the scope of this task. It depends on Particular/ProductionTests#33 and should be implemented after we figure out what performance tests should look like.

mauroservienti:
+1, we can be chickens if required.

tmasternak:
Created a new tasks for performance testing Particular/NServiceBus.SqlServer/issues/125

tmasternak:
This one is currently blocked until NHibernate is ready for wire-compat testing

PR #115, opened 2015-12-14
WIP: Compatibility Tests

petersgiles:
A working branch for developing tests

https://github.com/Particular/NServiceBus.SqlServer/issues/110

@weralabaj I figured we could work in here

### DIFF
--- a/src/NServiceBus.SqlServer.AcceptanceTests/NServiceBus.SqlServer.AcceptanceTests.csproj
+++ b/src/NServiceBus.SqlServer.AcceptanceTests/NServiceBus.SqlServer.AcceptanceTests.csproj
@@ -267,6 +267,7 @@
     <Compile Include="MultiSchema\When_custom_schema_configured.cs" />
     <Compile Include="MultiSchema\When_custom_schema_configured_with_transport_discriminator.cs" />
     <Compile Include="MultiSchema\When_custom_schema_configured_with_queue_specific_override.cs" />
+    <Compile Include="ServiceControl\When_an_endpoint_connects_to_multiple_databases.cs" />
     <Compile Include="When_using_custom_connection_factory.cs" />
   </ItemGroup>
   <ItemGroup>
@@ -275,6 +276,11 @@
       <Name>NServiceBus.SqlServer</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <Folder Include="Callbacks\" />
+    <Folder Include="Persistence\" />
+    <Folder Include="Scenarios\" />
+    <Folder Include="Transactions\" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
 </Project>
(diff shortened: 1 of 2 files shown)


weralabaj:
@petersgiles Cool, what are you working on?

petersgiles:
I was looking at service control compatibility. It's weird though. It seems that we've withdrawn multiple database support for NServiceBus.SqlServer (which Service Control doesn't support).

Also the tests that are currently in develop are all coming up red so I'm trying to work out what it going on there. Other than that I guess I'm going to start at the top and work down the list.

petersgiles:
Hmm the ServiceControl test I did probably doesn't belong in this repo maybe https://github.com/Particular/ProductionTests

weralabaj:
@petersgiles The support for multi-db in SqlTransport v3 (Core v6) is not removed, it's just not implemented yet ;)
It's WIP - /cc @mauroservienti @justabitofcode 

Regarding ServiceControl - you're talking about the issue with current implementation (desk case) or future plans? Either way would be good to sync!

petersgiles:
@SimonCropp was saying that a test in this repo for the configuration of SC is kind of daft (not his exact words) there's another mechanism for doing it when the product is built. I need him to repeat what it was because there were a lot of words about stuff i'm not familiar with

weralabaj:
OK, please, let us know how you decided to go about this then.

mauroservienti:
* I agree that we must test all the things that can interact/run in the ServiceControl process even if they are not specifically developed for ServiceControl
* I also thing that these tests don't belong here.

tmasternak:
@petersgiles any updates on that?

petersgiles:
@tmasternak well my feeling is that the tests should be moved to the core

SimonCropp:
How can tests that rely on sql transport be in the core?

tmasternak:
Do we have any SC integration tests?

mauroservienti:
@tmasternak not that I'm aware of. @johnsimons is there a way to run integration tests that involves an endpoint and SC?

petersgiles:
@SimonCropp many of these are actually NServiceBus integration tests and any transport would do. The file system transport that @andreasohlund was talking about putting into the core would be as appropriate. That's why I was thinking of the core.

If we are testing transport configuration then I would put that test in a transport solution.

SimonCropp:
@petersgiles i said it didnt make sense to have this in the sql server repo since it is testing multiple versions. a given point in a repo is a single instance. so u need to have it somewhere else where, gradually, more versions can be added to the compat matrix.

weralabaj:
@petersgiles Would it make sense to create SC tests in https://github.com/Particular/ProductionTests/ (see others here https://github.com/Particular/ProductionTests/tree/master/src)? Then we wouldn't limit it to the specific transport and its version.

petersgiles:
@weralabaj yes that makes sense to me

tmasternak:
Closing this one. I've opened issue in ProductionTests

weralabaj:
Btw @petersgiles if we could help with setting it up (me, Tomasz or Alessandro) let us know!

petersgiles:
go for it

PR #126, opened 2016-01-05
Support for all TransportTransactionModes

tmasternak:
@andreasohlund ready for review


src/NServiceBus.SqlServer/Receiving/ReceiveWithSendsAtomicWithReceiveTransaction.cs
```diff
@@ -0,0 +1,62 @@
...
+                {
+                    try
+                    {
+                        var readResult = await inputQueue.TryReceive(sqlConnection, transaction).ConfigureAwait(false);
+
+                        if (readResult.IsPoison)
+                        {
+                            await errorQueue.SendRawMessage(readResult.DataRecord, sqlConnection, transaction).ConfigureAwait(false);
+                        }
+                        else if (readResult.Successful)
```
andreasohlund: Nitpick: I like the early exit if there is a poison message better since that avoids the mental gymnastics of `else if` and reduces the nesting level?

tmasternak: The reason that I changed it is that we whould also need to commit transaction before return. In other works one line changes to three. TBH I'm not sure which option is better.

andreasohlund: >TBH I'm not sure which option is better.

I still like the early exit better even though it adds one line of code

MarcinHoppe:
:+1:

src/NServiceBus.SqlServer/Receiving/ReceiveWithSendsAtomicWithReceiveTransaction.cs
tmasternak: :+1:

tmasternak:
Feel free to merge when all green.

PR #144, opened 2016-01-26
ConfigureAwait(false)

danielmarbach:
@Particular/tf-async-transport please review

Also fixed the message pump. I think it is wrong to await the continuation task. You should be able to ignore the potentially synchronously executed continuation.


tmasternak:
:+1:

PR #151, opened 2016-02-03
[WIP] Switching to JsonConvert from ilmerged Json.Net for header serialization

tmasternak:

PR #150, opened 2016-02-03
[WIP] Remove reference to core jsonserializer

SzymonPobiega:

PR #152, opened 2016-02-04
add DictionarySerializer with DataContractJsonSerializer

SimonCropp:
@SzymonPobiega @tmasternak 

rather than ilmerging a 500K dll in the below how about using a DataContractJsonSerializer?
https://github.com/Particular/NServiceBus.SqlServer/pull/150
https://github.com/Particular/NServiceBus.SqlServer/pull/151

### DIFF
--- a/src/NServiceBus.SqlServer.AcceptanceTests/TransactionScope/When_customizing_scope_isolation_level.cs
+++ b/src/NServiceBus.SqlServer.AcceptanceTests/TransactionScope/When_customizing_scope_isolation_level.cs
@@ -22,18 +22,18 @@
             Assert.AreEqual(IsolationLevel.RepeatableRead, context.IsolationLevel, "Ambient transaction should have configured isolation level");
         }
 
-        private class MyMessage : IMessage
+        class MyMessage : IMessage
         {
         }
 
-        private class Context : ScenarioContext
+        class Context : ScenarioContext
         {
             public bool Done { get; set; }
             public bool AmbientTransactionPresent { get; set; }
             public IsolationLevel IsolationLevel { get; set; }
         }
 
-        private class Endpoint : EndpointConfigurationBuilder
+        class Endpoint : EndpointConfigurationBuilder
         {
             public Endpoint()
             {
--- a/src/NServiceBus.SqlServer.AcceptanceTests/TransactionScope/When_using_scope_timeout_greater_than_machine_max.cs
+++ b/src/NServiceBus.SqlServer.AcceptanceTests/TransactionScope/When_using_scope_timeout_greater_than_machine_max.cs
@@ -22,11 +22,11 @@
             Assert.That(exception.InnerException.InnerException.Message.Contains("Timeout requested is longer than the maximum value for this machine"));
         }
 
-        private class Context : ScenarioContext
+        class Context : ScenarioContext
         {
         }
 
-        private class Endpoint : EndpointConfigurationBuilder
+        class Endpoint : EndpointConfigurationBuilder
         {
             public Endpoint()
             {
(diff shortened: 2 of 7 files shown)


HEskandari:
I like it! :+1:

tmasternak:
@SimonCropp I think that argument was the time of serialization. I think that the difference is considerable. /cc: @danielmarbach

danielmarbach:
image

We feeled this matters

danielmarbach:
Just for your reference. We started using `DataContractSerializer` in CustomChecks and Heartbeats. After the performance discussion around SqlServer we decided to go with one way of doing serialization with Json (using JSon.NET) and bite the bullet of the bigger assemblies. How problematic is the size of the assembly anyway in that case? Can you elaborate?

SimonCropp:
in the scope of a a list of headers that are all strings is that performance comparison relevant?

tmasternak:
@SimonCropp I think we can't answer that without doing tests.

danielmarbach:
I agree with @tmasternak let's measure it.

tmasternak:
@SimonCropp @danielmarbach I'll make measurements and post it here.

SimonCropp:
@danielmarbach scratch the the perf testing. turns out DataContractJsonSerializer  and newtonsoft dont serialize a dictionary the same.

SimonCropp:
Json.net

`{"foo":"bar"}`

DataContractJsonSerializer
 
`[{"Key":"foo","Value":"bar"}]`

tmasternak:
http://stackoverflow.com/questions/4559991/any-way-to-make-datacontractjsonserializer-serialize-dictionaries-properly

tmasternak:
Ok let's make a step back here. Why is 500K so bad?

SimonCropp:
UseSimpleDictionaryFormat worked

danielmarbach:
@SimonCropp you can tweak it. We already use the emit option in the heartbeats https://github.com/Particular/ServiceControl.Plugin.Nsb6.Heartbeat/commit/93a892cec717eef7a3a017670a7e6a9fbc7f18a3#diff-f2f02cdc09cfc9465ca59cfa12954370R32

tmasternak:
Here are test results:

 | Assembly Size | WireCompat | Header De/Serialization | Message Delivery Snd/Rcv
------------ | -------------  | -------------  | -------------  | ------------- 
DataContractJsonSerializer | 62K + 0 | OK | S: 5660 ms  D: 16766 ms | S: 8000 msg/s R: 2000 msg/s S+R: 4000 msg/s, 850 msg/s
Json.Net |  62K + 500K | OK | S: 1851 ms D: 2770 ms | S: 8000 msg/s  R: 2000 msg/s S+R: 4000 msg/s, 1000 msg/s

3rd column -  shows times for serializing [S] and deserializing [D] sample Dictionary<string, string> for 100000 times
4th column -  shows throughput with only sender sending messages [S], only receiver receiving messages [R], and for sender and receiver running at the same time [S+R]

On my machine there is a difference in serialization/deserialization proces but there is no performance degradation when it comes to message throughput (probably because it is I/O bound).

I summary I propose (after syncing with @danielmarbach) to go with `DataContractJsonSerializer` and if in the future we see performance problems due to header serialization we can provide extension method for advanced users. 

/cc: @SimonCropp

SimonCropp:
Do we want to go with a different default encoding?

Scooletz:
@tmasternak Have you tried using [Jil](https://github.com/kevin-montrose/Jil)? If performance matters, there's nothing faster and more `low-alloc` in .NET world then this.

SzymonPobiega:
IMO a binary writer and type-length-value approach would be best but unfortunately not backwards wire-compatible. I think we are stuck with JSON :(

SimonCropp:
closing this one. did a clen PR here https://github.com/Particular/NServiceBus.SqlServer/pull/153

PR #163, opened 2016-02-11
Correctly remove expired messages from queue

MarcinHoppe:
SQL Server transport does not currently commit the transaction when it retrieves an expired message from the queue. This results in expired messages being processed indefinitely. This is clearly a regression from V2.

This PR addresses the problem in the same way as V2 (by committing the transaction).

### DIFF
--- a/src/NServiceBus.SqlServer.AcceptanceTests/TimeToBeReceived/When_queue_contains_expired_messages.cs
+++ b/src/NServiceBus.SqlServer.AcceptanceTests/TimeToBeReceived/When_queue_contains_expired_messages.cs
@@ -1,0 +1,86 @@
+﻿namespace NServiceBus.SqlServer.AcceptanceTests.TimeToBeReceived
+{
+    using System.Data.SqlClient;
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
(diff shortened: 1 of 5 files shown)


src/NServiceBus.SqlServer.AcceptanceTests/TimeToBeReceived/When_queue_contains_expired_messages.cs
```diff
@@ -0,0 +1,86 @@
...
+        class Context : ScenarioContext
+        {
+            public bool MessageWasHandled { get; set; }
+        }
+
+        class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>();
```
tmasternak: I think we need concurrecy set to 1 here otherwise we have race condition

MarcinHoppe: Good catch, fixed.

MarcinHoppe:
Build is green and the change was reviewed by @tmasternak. I think this is ready to merge.

MarcinHoppe:
@Particular/transport-maintainers Can you review and merge?

PR #158, opened 2016-02-12
V3: SQL Server transport now periodically purges expired messages

MarcinHoppe:

In the actual versions (V2 & V3) of the transport [TTBR messages](http://docs.particular.net/nservicebus/messaging/discard-old-messages) are handled by endpoint process, this means thjat if the input queue is load with a lot of expired messages the endpoint will waste a lot of time at startup dealing with expired messages before being able to start consuming non expired ones.

We should handle such a scenario better. /cc @tmasternak @ramonsmits 
Inception: https://github.com/Particular/CustomerSuccess/issues/166

### TODO:
- [x] Add index on the `Expires` column
- [x] Run performance tests
- [x] Doco PR
- [x] Backport to V2
- [x] Upgrade guide PR: Particular/docs.particular.net#1290

### DIFF
--- a/src/NServiceBus.SqlServer/Configuration/SettingsKeys.cs
+++ b/src/NServiceBus.SqlServer/Configuration/SettingsKeys.cs
@@ -10,7 +10,7 @@ namespace NServiceBus.Transports.SQLServer
 
         public const string LegacyMultiInstanceConnectionFactory = "SqlServer.LegacyMultiInstanceConnectionFactory";
 
-        public const string PurgeTaskDelayKey = "SqlServer.PurgeTaskDelay";
+        public const string PurgeTaskDelayTimeSpanKey = "SqlServer.PurgeTaskDelayTimeSpan";
         public const string PurgeBatchSizeKey = "SqlServer.PurgeBatchSize";
     }
 }
--- a/src/NServiceBus.SqlServer/Configuration/SettingsKeys.cs
+++ b/src/NServiceBus.SqlServer/Configuration/SettingsKeys.cs
@@ -9,5 +9,8 @@ namespace NServiceBus.Transports.SQLServer
         public const string TimeToWaitBeforeTriggering = "SqlServer.CircuitBreaker.TimeToWaitBeforeTriggering";
 
         public const string LegacyMultiInstanceConnectionFactory = "SqlServer.LegacyMultiInstanceConnectionFactory";
+
+        public const string PurgeTaskDelayKey = "SqlServer.PurgeTaskDelay";
+        public const string PurgeBatchSizeKey = "SqlServer.PurgeBatchSize";
     }
 }
--- a/src/NServiceBus.SqlServer/Queuing/Message.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Message.cs
@@ -14,7 +14,7 @@
             Headers = headers;
         }
 
-        public bool TTBRExpried(DateTime now)
+        public bool TTBRExpired(DateTime now)
         {
             return TimeToBeReceived.HasValue && TimeToBeReceived.Value < now;
         }
--- a/src/NServiceBus.SqlServer/Queuing/Message.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Message.cs
@@ -6,18 +6,18 @@
 
     class Message
     {
-        public Message(string transportId, DateTime? timeToBeReceived, Dictionary<string, string> headers, Stream bodyStream)
+        public Message(string transportId, int? millisecondsToExpiry, Dictionary<string, string> headers, Stream bodyStream)
         {
             TransportId = transportId;
-            TimeToBeReceived = timeToBeReceived;
+            if (millisecondsToExpiry.HasValue)
+            {
+                TimeToBeReceived = TimeSpan.FromMilliseconds(millisecondsToExpiry.Value);
+            }
             BodyStream = bodyStream;
             Headers = headers;
         }
 
-        public bool TTBRExpired(DateTime now)
-        {
-            return TimeToBeReceived.HasValue && TimeToBeReceived.Value < now;
-        }
+        public bool TTBRExpired => TimeToBeReceived.HasValue && TimeToBeReceived.Value.TotalMilliseconds < 0L;
 
         public string GetLogicalId()
         {
@@ -28,7 +28,7 @@
         }
 
         public string TransportId { get; }
-        public DateTime? TimeToBeReceived { get; }
+        public TimeSpan? TimeToBeReceived { get; }
         public Stream BodyStream { get; }
         public Dictionary<string, string> Headers { get; }
     }
--- a/src/NServiceBus.SqlServer/Queuing/Message.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Message.cs
@@ -1,6 +1,5 @@
 ﻿namespace NServiceBus.Transports.SQLServer
 {
-    using System;
     using System.Collections.Generic;
     using System.IO;
 
@@ -9,15 +8,10 @@
         public Message(string transportId, int? millisecondsToExpiry, Dictionary<string, string> headers, Stream bodyStream)
         {
             TransportId = transportId;
-            if (millisecondsToExpiry.HasValue)
-            {
-                TimeToBeReceived = TimeSpan.FromMilliseconds(millisecondsToExpiry.Value);
-            }
+            TTBRExpired = millisecondsToExpiry.HasValue && millisecondsToExpiry.Value < 0;
             BodyStream = bodyStream;
             Headers = headers;
         }
-
-        public bool TTBRExpired => TimeToBeReceived.HasValue && TimeToBeReceived.Value.TotalMilliseconds < 0L;
 
         public string GetLogicalId()
         {
@@ -28,7 +22,7 @@
         }
 
         public string TransportId { get; }
-        public TimeSpan? TimeToBeReceived { get; }
+        public bool TTBRExpired { get; }
         public Stream BodyStream { get; }
         public Dictionary<string, string> Headers { get; }
     }
--- a/src/NServiceBus.SqlServer/Queuing/Sql.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Sql.cs
@@ -44,6 +44,8 @@ namespace NServiceBus.Transports.SQLServer
                     EXEC sp_releaseapplock @Resource = '{0}_{1}_lock'
                   END";
 
+        internal const string PurgeBatchOfExpiredMessagesText = @"DELETE TOP({0}) FROM[{1}].[{2}] WHERE [Expires] < @Expires";
+
         internal class Columns
         {
             internal static ColumnInfo Id = new ColumnInfo
--- a/src/NServiceBus.SqlServer/Queuing/Sql.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Sql.cs
@@ -49,7 +49,7 @@ namespace NServiceBus.Transports.SQLServer
                     EXEC sp_releaseapplock @Resource = '{0}_{1}_lock'
                   END";
 
-        internal const string PurgeBatchOfExpiredMessagesText = @"DELETE TOP({0}) FROM[{1}].[{2}] WHERE [Expires] < @Expires";
+        internal const string PurgeBatchOfExpiredMessagesText = @"DELETE TOP({0}) FROM [{1}].[{2}] WITH (UPDLOCK, READPAST, ROWLOCK) WHERE [Expires] < @Expires";
 
         internal class Columns
         {
--- a/src/NServiceBus.SqlServer/Queuing/Sql.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Sql.cs
@@ -39,6 +39,11 @@ namespace NServiceBus.Transports.SQLServer
                         (
 	                        [RowVersion] ASC
                         )WITH (PAD_INDEX  = OFF, STATISTICS_NORECOMPUTE  = OFF, SORT_IN_TEMPDB = OFF, IGNORE_DUP_KEY = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS  = ON, ALLOW_PAGE_LOCKS  = ON) ON [PRIMARY]
+
+                        CREATE NONCLUSTERED INDEX [Index_Expires] ON [{0}].[{1}]
+                        (
+	                        [Expires] ASC
+                        )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON)
                     END
     
                     EXEC sp_releaseapplock @Resource = '{0}_{1}_lock'
--- a/src/NServiceBus.SqlServer/Queuing/Sql.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Sql.cs
@@ -8,13 +8,13 @@ namespace NServiceBus.Transports.SQLServer
 
         internal const string SendText =
             @"INSERT INTO [{0}].[{1}] ([Id],[CorrelationId],[ReplyToAddress],[Recoverable],[Expires],[Headers],[Body]) 
-                                    VALUES (@Id,@CorrelationId,@ReplyToAddress,@Recoverable,@Expires,@Headers,@Body)";
+                                    VALUES (@Id,@CorrelationId,@ReplyToAddress,@Recoverable,IIF(@TimeToBeReceivedMs IS NOT NULL, DATEADD(ms, @TimeToBeReceivedMs, GETUTCDATE()), NULL),@Headers,@Body)";
 
         internal const string ReceiveText =
             @"WITH message AS (SELECT TOP(1) * FROM [{0}].[{1}] WITH (UPDLOCK, READPAST, ROWLOCK) ORDER BY [RowVersion]) 
 			DELETE FROM message 
 			OUTPUT deleted.Id, deleted.CorrelationId, deleted.ReplyToAddress, 
-			deleted.Recoverable, deleted.Expires, deleted.Headers, deleted.Body;";
+			deleted.Recoverable, IIF(deleted.Expires IS NOT NULL, DATEDIFF(ms, GETUTCDATE(), deleted.Expires), NULL), deleted.Headers, deleted.Body;";
         
         internal const string PeekText = @"SELECT count(*) Id FROM [{0}].[{1}];";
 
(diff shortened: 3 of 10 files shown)


MarcinHoppe:
I took a first stab at solving the problem in V3 first.

The idea is to have a separate ongoing task in the message pump that tries to purge the expired messages _en masse_ in 10k row chunks (to make it easier on the database from locking and transaction log growth perspective). It is first fired upon the start of the message pump. Later on it will wake up every 5 minutes to try to purge expired messages. This is just a safeguard against the regular TTBR handling not catching up with incoming expired messages.

@mauroservienti @tmasternak @ramonsmits @andreasohlund @SzymonPobiega This is still WIP but feel free to review. The sooner we can find problems in this approach, the better.

tmasternak:
@MarcinHoppe I think that it would be great to make a test of purging during normal operation of the endpoint. E.g. We have 1000000 messages in the queue half of those are to be discarted due to TTBR. Receving and sending endpoints are up and running.

MarcinHoppe:
@tmasternak I ran some manual tests on a slightly smaller scale (about 300k TTBR-expired messages + some 30k non-expired messages) to check if the purging task does not choke the receiving endpoint (it did not).

I agree we would benefit from something more long-running to simulate a significant build-up of expired messages when the receiving endpoint is down. The challenge is to define what we will test for (e.g. impact on endpoint startup time, delay in getting to process non-expired messages, etc.).

Thoughts?

mauroservienti:
not only that, we should take into account (when testing) latency and I/O, would be interesting to test using a VM based environment where we can tweak the VM performance, making them worse than average and observe what happens.

MarcinHoppe:
I think the plan of attack on this would be to set up a test environment in the cloud (AWS or Azure), run a mixed workload (TTBR + non-TTBR) messages where both endpoints are running and establish a performance baseline over a set of metrics (to be defined), then shut down the receiving endpoint, let the queue grow, then start the receiving endpoint and gather those same metrics again.

@tmasternak @mauroservienti I think a reasonable set of metrics is to begin with are the standard NSB performance counters + endpoint start time.

Thoughts?

Potentially there's some overlap with Particular/ProductionTests#46 in terms of infrastructure needed. //cc @WojcikMike

tmasternak:
I like the idea. I think that some real world values in terms of load and TTBR should come from service control client cases.

MarcinHoppe:
The 300k volume that I used is taken from the LightSail case. The TTBR could be the default value used by the SC heartbeats plugin for the heartbeat TTL.

mauroservienti:
Local Hyper-V on Windows allows to tweak VM performance settings, so there is no need to setup a cloud environment unless is required for other purposes.

MarcinHoppe:
@mauroservienti This might a good first shot before may have an opportunity to piggy-back on the infrastructure for Outbox performance testing.

MarcinHoppe:
I managed to get some measurements on the performance impact of this feature on receiving messages.

### What was tested?

I pumped the database with 500k messages: 250k expired and 250k w/o TTBR. Both types of messages were interwined. Then I started an endpoint with empty handlers and let it drain 100k messages from the queue. I measured the throughput of the endpoint using standard NSB performance counters and the time it takes to purge all the expired messages at startup.


### Results

- In baseline scenario about half the throughput is spent on handling expired messages.
- In the improved case the endpoint __starts to process non-expired messages immediately__ while the purge task removes expired messages in the background.
	- If head of the queue had a huge backlog of expired messages then the endpoint would still process some expired messages before sufficient number of chunks have been removed from the head of the queue.
- Time to purge a realistic volume of expired messages on endpoint startup seems to be acceptable.
- Performance would still be acceptable for users that only update the SQL transport NuGet but not create the index.
 

### Measurements

Metric|Result|Comment
------|------|-------
Baseline (no fix)|77 msg/s|Endpoint processes both expired and non-expired messages
Purge on startup (with index)|141 msg/s|Endpoint processes only non-expired messages
Time to purge all expires messages (with index)|14,5s|Purge happens in parallel with processing
Purge on startup (without index)|135 msg/s|Endpoint processes only non-expired messages
Time to purge all expires messages (without index)|22,5s|Purge happens in parallel with processing


### Setup

All the tests were executed on my laptop:

- CPU: Intel Core i7-6700HQ @ 2.60 MHz
- RAM: 16 GB RAM
- SSD: 500 GB
- OS: Windows 10 Pro

I created a client Hyper-V VM for SQL Server:

- CPU: 4 virtual CPUs
- RAM: 8 GB (static)
- Drives (all static):
	- OS (32 GB)
	- Data (8 GB)
	- Log (4 GB)
- OS: Windows Server 2012 R2 Standard
- Database: SQL Server 2014 Enterprise Edition

The receving endpoint was running on the host. The sending endpoint was only used to fill up the database and was not running during the test.

src/NServiceBus.SqlServer/Queuing/Sql.cs
```diff
@@ -39,11 +39,18 @@ class Sql
...
+                        CREATE NONCLUSTERED INDEX [Index_Expires] ON [{0}].[{1}]
+                        (
+	                        [Expires] ASC
+                        )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON)
                     END
     
                     EXEC sp_releaseapplock @Resource = '{0}_{1}_lock'
                   END";
 
+        internal const string PurgeBatchOfExpiredMessagesText = @"DELETE TOP({0}) FROM [{1}].[{2}] WITH (UPDLOCK, READPAST, ROWLOCK) WHERE [Expires] < @Expires";
```
tmasternak: I would change the @Expires to @CurrentTime to make it more explicit what this parameter does

mauroservienti:
@MarcinHoppe great job, kudos.

> Performance would still be acceptable for users that only update the SQL transport NuGet but not create the index.

Do we have an option to log a warning, at startup only, if the index is not there?

MarcinHoppe:
@mauroservienti I guess it comes down to permissions: I need to verify if a user that has the permissions to query the queue table can also query for metadata about this table. Stay tuned.

At a minimum, we could do a best effort check on endpoint startup and silently fail if there's a permission problem.

mauroservienti:
Good point @MarcinHoppe, that's why several month ago I raised this: https://github.com/Particular/PlatformDevelopment/issues/202

ramonsmits:
@MarcinHoppe Strange that DELETE query takes so long. Did a similar query on a customer production database containing 500.000+ messages and each 'batch' of 10.000 ran a few milliseconds.

MarcinHoppe:
@ramonsmits It'd have to do more profiling that is is that actually takes up the time. Did you do a loop in the code sending query one by one or was the entire loop in T-SQL? Maybe here it is a better idea to send the entire T-SQL loop as a single query from ADO.NET perspective? I'll experiment on the V2 hotfix branch.

ramonsmits:
@MarcinHoppe The query ran within the SQL Manager in a few rows.

That could be simulated by running that SQL statement comparable to

DELETE TOP (10000 FROM X
GO
DELETE TOP (10000 FROM X
GO
DELETE TOP (10000 FROM X
GO
DELETE TOP (10000 FROM X
GO

That would be one SQL roundtrip but I think that is not needed.

MarcinHoppe:
@ramonsmits My test had to purge 250k expired messages. With 10k batch size this was 25 roundtrips. Look like a lot and potentially encoding a loop and having it run in one roundtrip should be much more efficient.

@mauroservienti @tmasternak Thoughts?

mauroservienti:
Which are the differences between this and #165? I'm confused :smile:

MarcinHoppe:
@mauroservienti V2 vs V3

mauroservienti:
ooops, got it, thanks

mauroservienti:
@MarcinHoppe @Particular/platform-dev-squad this is missing a proper task force and as discussed should be connected to an issue and not a PR on its own.

MarcinHoppe:
Shouldn't this be added to the 3.0.0 milestone? Tagged as "Core v6"?

mauroservienti:
oh yes, done

MarcinHoppe:
@Particular/transport-maintainers This is rebased and has "forward"-ported features from V2 implementation of the fix.

Please review and merge if OK.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,94 @@
...
+    using NServiceBus.Settings;
+
+    interface IPurgeExpiredMessages
+    {
+        Task Purge(TableBasedQueue queue, CancellationToken cancellationToken);
+    }
+
+    class ExpiredMessagesPurger : IPurgeExpiredMessages
+    {
+        public ExpiredMessagesPurger(SettingsHolder settings, Func<TableBasedQueue, Task<SqlConnection>> openConnection)
```
tmasternak: Can we pass the settings values explicitly? We do not need to rely on the whole `SettingsHolder`.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,94 @@
+﻿namespace NServiceBus.Transports.SQLServer
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using NServiceBus.Logging;
+    using NServiceBus.Settings;
+
+    interface IPurgeExpiredMessages
```
tmasternak: Why do we need this interface?

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,94 @@
...
+                    Logger.InfoFormat("{0} expired messages were successfully purged from table {1}", totalPurgedRowsCount, queue);
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.WarnFormat("Purging expired messages from table {0} failed after purging {1} messages. Exception: {2}", queue, totalPurgedRowsCount, ex);
+            }
+
+            Logger.DebugFormat("Scheduling next expired message purge task for table {0} in {1}", queue, purgeTaskDelay);
+            await Task.Delay(purgeTaskDelay, cancellationToken).ConfigureAwait(false);
```
tmasternak: I think this should be in the calling/scheduling method.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,94 @@
...
+                    while (continuePurging && !cancellationToken.IsCancellationRequested)
+                    {
+                        var purgedRowsCount = await queue.PurgeBatchOfExpiredMessages(connection, purgeBatchSize).ConfigureAwait(false);
+
+                        totalPurgedRowsCount += purgedRowsCount;
+                        continuePurging = (purgedRowsCount == purgeBatchSize);
+                    }
+                }
+
+                if (totalPurgedRowsCount > 0)
```
tmasternak: I think that for production debugging purposes we should log a line on each purge end no matter how many messages were deleted.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            await LogWarningWhenIndexIsMissing(queue).ConfigureAwait(false);
+
+            Logger.DebugFormat("Starting a new expired message purge task for table {0}.", queue);
+
+            var totalPurgedRowsCount = 0;
+
+            try
+            {
+                using (var connection = await openConnection(queue).ConfigureAwait(false))
```
tmasternak: Why not use the same connection for logging warnings and purging messages?

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    var continuePurging = true;
+
+                    while (continuePurging && !cancellationToken.IsCancellationRequested)
```
tmasternak: Why not `do-while`?

src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
```diff
@@ -133,6 +133,37 @@ public Task SendMessage(OutgoingMessage message, SqlConnection connection, SqlTr
...
+        {
+            var commandText = string.Format(Sql.CheckIfExpiresIndexIsPresent, Sql.ExpiresIndexName, this.address.SchemaName, this.address.TableName);
+
+            using (var command = new SqlCommand(commandText, connection))
+            {
+                var rowsCount = (int) await command.ExecuteScalarAsync().ConfigureAwait(false);
+
+                if (rowsCount == 0)
+                {
+                    Logger.WarnFormat(@"Table [{0}].[{1}] does not contain index '{2}'.
```
tmasternak: Can we reformat this? Now the line wraps because of the long string. I would use `+` to break the string in two.

src/NServiceBus.SqlServer/Configuration/SettingsKeys.cs
```diff
@@ -9,5 +9,8 @@ class SettingsKeys
         public const string TimeToWaitBeforeTriggering = "SqlServer.CircuitBreaker.TimeToWaitBeforeTriggering";
 
         public const string LegacyMultiInstanceConnectionFactory = "SqlServer.LegacyMultiInstanceConnectionFactory";
+
+        public const string PurgeTaskDelayKey = "SqlServer.PurgeTaskDelay";
```
ramonsmits: As these contains strings, mention what the value resembles. If is seconds, milliseconds, hours, timespan?

src/NServiceBus.SqlServer/Queuing/Message.cs
```diff
@@ -28,7 +28,7 @@ public string GetLogicalId()
         }
 
         public string TransportId { get; }
-        public DateTime? TimeToBeReceived { get; }
+        public TimeSpan? TimeToBeReceived { get; }
```
tmasternak: This is used only internally.

src/NServiceBus.SqlServer/Queuing/Sql.cs
```diff
@@ -8,13 +8,13 @@ class Sql
 
         internal const string SendText =
             @"INSERT INTO [{0}].[{1}] ([Id],[CorrelationId],[ReplyToAddress],[Recoverable],[Expires],[Headers],[Body]) 
-                                    VALUES (@Id,@CorrelationId,@ReplyToAddress,@Recoverable,@Expires,@Headers,@Body)";
+                                    VALUES (@Id,@CorrelationId,@ReplyToAddress,@Recoverable,IIF(@TimeToBeReceivedMs IS NOT NULL, DATEADD(ms, @TimeToBeReceivedMs, GETUTCDATE()), NULL),@Headers,@Body)";
```
ramonsmits: Neat server side clock solution.

src/NServiceBus.SqlServer/Queuing/Message.cs
```diff
@@ -6,18 +6,18 @@
 
     class Message
     {
-        public Message(string transportId, DateTime? timeToBeReceived, Dictionary<string, string> headers, Stream bodyStream)
+        public Message(string transportId, int? millisecondsToExpiry, Dictionary<string, string> headers, Stream bodyStream)
```
tmasternak: Can we pass `TimeSpan` value here. and calculate `TTBRExpired` here as well? That will allow us to remove `TimeToBeReceived` variable altogether.

src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
ramonsmits: or use \n ?

src/NServiceBus.SqlServer/Queuing/Sql.cs
tmasternak: :beer:

tmasternak:
I've done my review.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
ramonsmits: I don't like that this is a INFO. It is expected that expired message are purged. If this is about metrics then it should be DEBUG, can we maybe also include the duration?

src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
```diff
@@ -133,6 +133,37 @@ public Task SendMessage(OutgoingMessage message, SqlConnection connection, SqlTr
 
         public string TransportAddress => address.ToString();
 
+        public async Task<int> PurgeBatchOfExpiredMessages(SqlConnection connection, int purgeBatchSize)
+        {
+            var commandText = string.Format(Sql.PurgeBatchOfExpiredMessagesText, purgeBatchSize, address.SchemaName, address.TableName);
+
+            using (var command = new SqlCommand(commandText, connection))
+            {
+                return await command.ExecuteNonQueryAsync().ConfigureAwait(false);
```
tmasternak: No need for `async` method and `await` we can just return `Task.Of(T)`

MarcinHoppe: @tmasternak I think `await` (and hence `async`) is needed to call `Dispose` on the `command` in the continuation.

tmasternak: :+1:  my bad

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
MarcinHoppe: @tmasternak This would expose details of the purger (like delay) to the outside world. What would be the benefit?

`QueuePeeker` uses the same design (i.e. managed the delay internally).

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
MarcinHoppe: @tmasternak `while` loop allows to roll initial (before the end of the loop) check for cancellation request into the loop condition itself. `do-while` approach was more complicated compared to this version (more lines of code and more control structures).

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
tmasternak: :+1: Fair enough. We can address both of them during refactorings.

src/NServiceBus.SqlServer/Configuration/SettingsKeys.cs
MarcinHoppe: Fixed.

src/NServiceBus.SqlServer/Queuing/Message.cs
MarcinHoppe: Fixed.

src/NServiceBus.SqlServer/Queuing/Message.cs
MarcinHoppe: Fixed.

src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
MarcinHoppe: Fixed.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
MarcinHoppe: I thought I would be implementing a legacy version of the purger. I removed this interface.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
MarcinHoppe: Fixed.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
MarcinHoppe: Not applicable anymore since I split index check and purging into separate public methods called separately from the message pump.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
MarcinHoppe: Fixed.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
MarcinHoppe: Fixed it anyways :smile:.

MarcinHoppe:
@Particular/transport-maintainers All comments have been addressed. Ready to merge.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
tmasternak: :beers:

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
tmasternak: :+1:

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
tmasternak: :+1:

adamralph:
> as discussed should be connected to an issue and not a PR on its own.

please see https://github.com/Particular/NServiceBus.SqlServer/pull/165#issuecomment-195351291

PR #165, opened 2016-02-22
V2: SQL Server transport now periodically purges expired messages

MarcinHoppe:

SQL Server transport discards expired messages when it receives them from the queue table. If the endpoint is down for some time and expired messages build up in the queue, the endpoint will need to receive all of these messages in order to discard them.

The solution is to purge expired messages at the startup of the endpoint and then periodically after every 5 minutes. The purging is done in batches, 10000 rows at a time until there are no more expired messages in the queue.

## Who's affected

Users of the SQL Server transport receiving messages with `TimeToBeReceived` attribute.

## Symptoms

When there is a backlog of expired messages in the queue table, then upon the startup of the endpoint the log will contain the following line for each expired message:

`Message with ID=<Message ID> has expired. Removing it from queue.`

## Deployment

Optimal performance of this solution requires that an additional index be created on each of the queue tables. The following SQL statement may be used to create the index:

```SQL
CREATE NONCLUSTERED INDEX [Index_Expires] ON [<SCHEMA>].[<QUEUE_TABLE>]
(
	[Expires] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON)
```
## Workaround

A manual workaround is to manually purge expired messages using the following SQL statement:

```SQL
DELETE FROM [<SCHEMA>].[<QUEUE_TABLE>] WITH (UPDLOCK, READPAST, ROWLOCK)
WHERE [Expires] < GETUTCDATE()
```

The query above may cause a significant growth of the transaction log if there are many rows affected. The recommended technique to mitigate this problem is to split the single query into chunks:

```SQL
DECLARE @rows_deleted INT = 1;

WHILE @rows_deleted > 0
BEGIN
	BEGIN TRANSACTION;

	DELETE TOP(10000) FROM [<SCHEMA>].[<QUEUE_TABLE>] WITH (UPDLOCK, READPAST, ROWLOCK)
	WHERE [Expires] < GETUTCDATE();

	SET @rows_deleted = @@ROWCOUNT;

	COMMIT TRANSACTION;
END
```
### Things to be completed before release

- [x] Doco PR: Particular/docs.particular.net#1224
- [x] Announcement PR: Particular/Announcements#82

### Things to be completed after release
- [ ] Ping the @Particular/servicecontrol-maintainers to schedule a ServiceControl release

### DIFF
--- a/src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
+++ b/src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
@@ -1,0 +1,120 @@
+﻿namespace NServiceBus.Transports.SQLServer
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Threading;
+    using NServiceBus.Logging;
+
+    class ExpiredMessagesPurger : IExecutor
+    {
+        static readonly TimeSpan purgeTaskDelay = TimeSpan.FromMinutes(5);
+
+        static readonly ILog Logger = LogManager.GetLogger(typeof(ExpiredMessagesPurger));
+
+        readonly TableBasedQueue queue;
+        readonly Func<SqlConnection> openConnection;
+
+        Timer purgeTaskTimer;
+        CancellationToken token;
+
+        object lockObject = new object();
+
+        public ExpiredMessagesPurger(TableBasedQueue queue, Func<SqlConnection> openConnection)
+        {
+            this.queue = queue;
+            this.openConnection = openConnection;
+        }
+
+        public void Start(int maximumConcurrency, CancellationToken token)
+        {
+            LogWarningWhenIndexIsMissing();
+
+            this.token = token;
+            purgeTaskTimer = new Timer(PurgeExpiredMessagesCallback, null, TimeSpan.Zero, purgeTaskDelay);
+        }
+
+        public void Stop()
+        {
+            using (var waitHandle = new ManualResetEvent(false))
+            {
+                purgeTaskTimer.Dispose(waitHandle);
+                waitHandle.WaitOne();
+            }
+        }
+
+        void LogWarningWhenIndexIsMissing()
+        {
+            try
+            {
+                using (var connection = openConnection())
+                {
+                    queue.LogWarningWhenIndexIsMissing(connection);
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.WarnFormat("Checking indexes on table {0} failed. Exception: {1}", queue, ex);
+            }
+        }
+
+        void PurgeExpiredMessagesCallback(object state)
+        {
+            Logger.DebugFormat("Starting a new expired message purge task for table {0}.", queue);
+
+            if (token.IsCancellationRequested)
+            {
+                return;
+            }
+
+            var lockTaken = false;
+            try
+            {
+                Monitor.TryEnter(lockObject, ref lockTaken);
+                if (!lockTaken)
+                {
+                    Logger.DebugFormat("An expired message purge task for table {0} is already running. Nothing to do.", queue);
+                    return;
+                }
+
+                PurgeExpiredMessages();
+            }
+            finally
+            {
+                if (lockTaken)
+                {
+                    Monitor.Exit(lockObject);
+                }
+            }
+        }
+
+        void PurgeExpiredMessages()
+        {
+            int totalPurgedRowsCount = 0;
+
+            try
+            {
+                using (var connection = openConnection())
+                {
+                    var continuePurging = true;
+
+                    while (continuePurging && !token.IsCancellationRequested)
+                    {
+                        var purgedRowsCount = queue.PurgeBatchOfExpiredMessages(connection);
+
+                        totalPurgedRowsCount += purgedRowsCount;
+                        continuePurging = (purgedRowsCount == TableBasedQueue.PurgeBatchSize);
+                    }
+                }
+
+                if (totalPurgedRowsCount > 0)
+                {
+                    Logger.InfoFormat("{0} expired messages were successfully purged from table {1}", totalPurgedRowsCount, queue);
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.WarnFormat("Purging expired messages from table {0} failed after purging {1} messages. Exception: {2}", queue, totalPurgedRowsCount, ex);
+            }
+        }
+    }
+}
--- a/src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
+++ b/src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
@@ -7,22 +7,20 @@
 
     class ExpiredMessagesPurger : IExecutor
     {
-        static readonly TimeSpan purgeTaskDelay = TimeSpan.FromMinutes(5);
-
-        static readonly ILog Logger = LogManager.GetLogger(typeof(ExpiredMessagesPurger));
+ static readonly ILog Logger = LogManager.GetLogger(typeof(ExpiredMessagesPurger));
 
         readonly TableBasedQueue queue;
         readonly Func<SqlConnection> openConnection;
+        readonly PurgeExpiredMessagesParams parameters;
 
         Timer purgeTaskTimer;
         CancellationToken token;
 
-        object lockObject = new object();
-
-        public ExpiredMessagesPurger(TableBasedQueue queue, Func<SqlConnection> openConnection)
+        public ExpiredMessagesPurger(TableBasedQueue queue, Func<SqlConnection> openConnection, PurgeExpiredMessagesParams parameters)
         {
             this.queue = queue;
             this.openConnection = openConnection;
+            this.parameters = parameters;
         }
 
         public void Start(int maximumConcurrency, CancellationToken token)
@@ -30,7 +28,7 @@
             LogWarningWhenIndexIsMissing();
 
             this.token = token;
-            purgeTaskTimer = new Timer(PurgeExpiredMessagesCallback, null, TimeSpan.Zero, purgeTaskDelay);
+            purgeTaskTimer = new Timer(PurgeExpiredMessagesCallback, null, TimeSpan.Zero, Timeout.InfiniteTimeSpan);
         }
 
         public void Stop()
@@ -66,25 +64,15 @@
                 return;
             }
 
-            var lockTaken = false;
-            try
-            {
-                Monitor.TryEnter(lockObject, ref lockTaken);
-                if (!lockTaken)
-                {
-                    Logger.DebugFormat("An expired message purge task for table {0} is already running. Nothing to do.", queue);
-                    return;
-                }
+            PurgeExpiredMessages();
 
-                PurgeExpiredMessages();
-            }
-            finally
+            if (token.IsCancellationRequested)
             {
-                if (lockTaken)
-                {
-                    Monitor.Exit(lockObject);
-                }
+                return;
             }
+
+            Logger.DebugFormat("Scheduling next expired message purge task for table {0} in {1}", queue, parameters.PurgeTaskDelay);
+            purgeTaskTimer.Change(parameters.PurgeTaskDelay, Timeout.InfiniteTimeSpan);
         }
 
         void PurgeExpiredMessages()
@@ -99,10 +87,10 @@
 
                     while (continuePurging && !token.IsCancellationRequested)
                     {
-                        var purgedRowsCount = queue.PurgeBatchOfExpiredMessages(connection);
+                        var purgedRowsCount = queue.PurgeBatchOfExpiredMessages(connection, parameters.PurgeBatchSize);
 
                         totalPurgedRowsCount += purgedRowsCount;
-                        continuePurging = (purgedRowsCount == TableBasedQueue.PurgeBatchSize);
+                        continuePurging = (purgedRowsCount == parameters.PurgeBatchSize);
                     }
                 }
 
--- a/src/NServiceBus.SqlServer/SqlServerQueueCreator.cs
+++ b/src/NServiceBus.SqlServer/SqlServerQueueCreator.cs
@@ -27,7 +27,13 @@ namespace NServiceBus.Transports.SQLServer
                     CREATE NONCLUSTERED INDEX [Index_Expires] ON [{0}].[{1}]
                     (
 	                    [Expires] ASC
-                    )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON)
+                    )
+                    INCLUDE
+                    (
+                        [Id],
+                        [RowVersion]
+                    )
+                    WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON)
 
                   END";
 
--- a/src/NServiceBus.SqlServer/SqlServerQueueCreator.cs
+++ b/src/NServiceBus.SqlServer/SqlServerQueueCreator.cs
@@ -24,6 +24,11 @@ namespace NServiceBus.Transports.SQLServer
 	                    [RowVersion] ASC
                     )WITH (PAD_INDEX  = OFF, STATISTICS_NORECOMPUTE  = OFF, SORT_IN_TEMPDB = OFF, IGNORE_DUP_KEY = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS  = ON, ALLOW_PAGE_LOCKS  = ON) ON [PRIMARY]
                     
+                    CREATE NONCLUSTERED INDEX [Index_Expires] ON [{0}].[{1}]
+                    (
+	                    [Expires] ASC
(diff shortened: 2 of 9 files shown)


MarcinHoppe:
@Particular/transport-maintainers Please review (please note items listed in the TODO section are WIP).

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,95 @@
...
+
+                        totalPurgedRowsCount += purgedRowsCount;
+                    } while (purgedRowsCount == TableBasedQueue.PurgeBatchSize);
+                }
+
+                Logger.Info($"{totalPurgedRowsCount} expired messages were successfully purged from table {queue}");
+            }
+            catch (Exception ex)
+            {
+                Logger.Warn($"Purging expired messages from table {queue} failed after purging {totalPurgedRowsCount} messages.", ex);
```
SimonCropp: Why would we not error here?

src/NServiceBus.SqlServer/SqlServerQueueCreator.cs
```diff
@@ -24,6 +24,11 @@ class SqlServerQueueCreator : ICreateQueues
 	                    [RowVersion] ASC
                     )WITH (PAD_INDEX  = OFF, STATISTICS_NORECOMPUTE  = OFF, SORT_IN_TEMPDB = OFF, IGNORE_DUP_KEY = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS  = ON, ALLOW_PAGE_LOCKS  = ON) ON [PRIMARY]
                     
+                    CREATE NONCLUSTERED INDEX [Index_Expires] ON [{0}].[{1}]
```
SimonCropp: How will this be run for existing environments?

src/NServiceBus.SqlServer/TableBasedQueue.cs
```diff
@@ -189,6 +189,36 @@ static T GetNullableValue<T>(object value)
...
 
+        public int PurgeBatchOfExpiredMessages(SqlConnection connection)
+        {
+            var commandText = string.Format(SqlPurgeBatchOfExpiredMessages, PurgeBatchSize, this.schema, this.tableName);
+
+            using (var command = new SqlCommand(commandText, connection))
+            {
+                command.Parameters.Add("UTCNow", SqlDbType.DateTime).Value = DateTime.UtcNow;
+
+                var rowsCount = command.ExecuteNonQuery();
```
SimonCropp: Just return

src/NServiceBus.SqlServer/TableBasedQueue.cs
```diff
@@ -189,6 +189,36 @@ static T GetNullableValue<T>(object value)
...
+
+                return rowsCount;
+            }
+        }
+
+        public void LogWarningWhenIndexIsMissing(SqlConnection connection)
+        {
+            var commandText = string.Format(SqlCheckIfExpiresIndexIsPresent, ExpiresIndexName, this.schema, this.tableName);
```
SimonCropp: Are there any possible permission concerns here? I assume earlier versions would operate ok without permissions to sys tables?

src/NServiceBus.SqlServer/SqlServerQueueCreator.cs
mauroservienti: unfortunately there isn't much we can do other than try to understand if the index is there and if not log a warning saying that we need it. Any suggestion on how this can be improved @SimonCropp?
Obviously we'll have documentation on this, in the migration guide as well.

src/NServiceBus.SqlServer/TableBasedQueue.cs
```diff
@@ -189,6 +189,36 @@ static T GetNullableValue<T>(object value)
...
+
+            using (var command = new SqlCommand(commandText, connection))
+            {
+                var rowsCount = (int) command.ExecuteScalar();
+
+                if (rowsCount == 0)
+                {
+                    Logger.Warn($@"Table [{schema}].[{tableName}] does not contain index '{ExpiresIndexName}'.
```
SimonCropp: any reasons this shouldnt be an error?

src/NServiceBus.SqlServer/SqlServerQueueCreator.cs
SimonCropp: i actually think is address adequately below. although i do have concerns about the permissions aspect https://github.com/Particular/NServiceBus.SqlServer/pull/165/files#r53696419

mauroservienti: If I recall correctly @MarcinHoppe tested it using a low privileged SQL login, but I'll let him comment with the research results.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
MarcinHoppe: @SimonCropp My thinking was that failure to purge is not fatal, i.e. the endpoint can still process incoming messages. Messages with expired TTBR will still be discarded upon receive.

src/NServiceBus.SqlServer/TableBasedQueue.cs
MarcinHoppe: @SimonCropp The purging procedure will still work without the index present but the operation will be slower. Here are some benchmarks for V3 implementation of this feature: https://github.com/Particular/NServiceBus.SqlServer/pull/158#issuecomment-185379258.

src/NServiceBus.SqlServer/TableBasedQueue.cs
MarcinHoppe: @SimonCropp It is sufficient to have `GRANT SELECT` permission on the table to be able to query metadata about this table (https://msdn.microsoft.com/en-us/library/ms173760.aspx#Anchor_0). I actually checked this in practice.

src/NServiceBus.SqlServer/SqlServerQueueCreator.cs
MarcinHoppe: @SimonCropp @mauroservienti Yep, I did check it with a low privileged account. See https://github.com/Particular/NServiceBus.SqlServer/pull/165/files#r53750491.

SimonCropp: Ok awesome

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,95 @@
...
+        static readonly ILog Logger = LogManager.GetLogger(typeof(ExpiredMessagesPurger));
+        readonly ConnectionFactory connectionFactory;
+        readonly LocalConnectionParams localConnectionParams;
+
+        readonly TableBasedQueue queue;
+
+        Timer purgeTaskTimer;
+        CancellationToken token;
+
+        public ExpiredMessagesPurger(TableBasedQueue queue, LocalConnectionParams localConnectionParams, ConnectionFactory connectionFactory)
```
tmasternak: Can we pass only connectionString instead of localConnectionParams? Or single factory method instead of localConnectionParams and connectionFactory?

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            try
+            {
+                using (var connection = connectionFactory.OpenNewConnection(localConnectionParams.ConnectionString))
+                {
+                    int purgedRowsCount;
+
+                    do
+                    {
+                        if (token.IsCancellationRequested)
```
tmasternak: I would combine that with condition in while loop.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,95 @@
...
+                            break;
+                        }
+
+                        purgedRowsCount = queue.PurgeBatchOfExpiredMessages(connection);
+
+                        totalPurgedRowsCount += purgedRowsCount;
+                    } while (purgedRowsCount == TableBasedQueue.PurgeBatchSize);
+                }
+
+                Logger.Info($"{totalPurgedRowsCount} expired messages were successfully purged from table {queue}");
```
tmasternak: Previously each purged message would have separate [log entry](https://github.com/Particular/NServiceBus.SqlServer/blob/hotfix-2.2.2/src/NServiceBus.SqlServer/TableBasedQueue.cs#L141) with message id. Do we care to put that in the release notes?

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
tmasternak: I think that in [other parts](https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.SqlServer/AdaptiveExecutor.cs#L79) we are quite conservative about that. If we have problems with transport operations we fail-fast.

MarcinHoppe: @tmasternak The reason the purger does not trigger the circuit breaker is that it could e.g. timeout without causing the endpoint to stop processing messages (which would require a fail-fast approach). If there's a database connection problem, we will know about it much faster from the receive loop.

Does this make sense?

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
MarcinHoppe: @tmasternak It still will do it if an expired message is received between periodic runs of the purger. This behaviour remains intact. What changes, though is that previously each and every expired message ID was in the log, now it won't.

Do other transports log IDs of expired messages? If not, then I would not bother.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
tmasternak: I agree that receive loop will shutdown the endpoint faster most of the times. That being said I would argue that the endpoint should go down if we can receive messages but not purge expired for some reason. In other words I think that we should be conservative here. Purger can do it's work => endpoint is not working correctly => we should shutdown.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
tmasternak: @MarcinHoppe I think you are right. I don't think other transports are doing that. And I don't consider logger entries as api user should depend on. We have notifications for those kind of things.

MarcinHoppe: :+1:

src/NServiceBus.SqlServer/TableBasedQueue.cs
MarcinHoppe: Fixed.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
MarcinHoppe: Good idea, looks better now.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
MarcinHoppe: Fixed.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
MarcinHoppe: @tmasternak You want to shut down the endpoint altogether or just arm the circuit breaker?

tmasternak: @MarcinHoppe I was thinking about arming circuit breaker but in the scenario that I have described it you be disarmed by the receive loop. In other words I think that supporting scenario I've described we would have to have a separate circuit breaker for purger.

tmasternak: @MarcinHoppe I looks that we would need to have a separate circuit breaker for purger to support scenario that I've described. I don't think it is necessary at the moment.

ramonsmits:
Regarding the use of `DateTime.UtcNow` instead of `GETUTCDATE()`. Considering clock drift I think still `GETUTCDATE()` should be used. Its the most consistent timestamp from a data perspective. What when we have multiple consumers where each has a different clock drift?

If that really is an issue then we should include `WHERE [Expires] IS NULL OR Expires>GETUTCDATE()` in the select query.

We could even allow for a small clock drift correction on the consumer by comparing with `DateTime.UtcNow.Add((TimeSpan)AllowedClockDrift)`

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,90 @@
...
+            this.queue = queue;
+            this.openConnection = openConnection;
+        }
+
+        public void Start(int maximumConcurrency, CancellationToken token)
+        {
+            LogWarningWhenIndexIsMissing();
+
+            this.token = token;
+            purgeTaskTimer = new Timer(_ => PurgeExpiredMessages(), null, TimeSpan.Zero, purgeTaskDelay);
```
ramonsmits: Call me old fashioned but I prefer non-lamba usage and naming `PurgeExpiredMessages` to indicate its invoked by a timer.

MarcinHoppe:
@ramonsmits The competing consumer scenario is a good argument. Now I'm thinking we should use `GETUTCDATE()` both when sending the message and when making an expiration decision.

@Particular/transport-maintainers Any thoughts? I'm :+1: for making the switch in this PR.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
MarcinHoppe: We like old fashioned. Fixed.

mauroservienti:
I'm :+1: too to switch to `GETUTCDATE()`, as well. @ramonsmits raised a very good point.

MarcinHoppe:
@mauroservienti @ramonsmits All is good with `GETUTCDATE()` for send and purge expired operations. However, during regular receive the discarding of expired messages also relies on `DateTime.UtcNow`.

What would be the preferred option to solve this one?

I'm thinking that the DELETE statement that does the receive should return a diff between `Expires` column and  `GETUTCDATE()` (the receive moment) and based on this diff check if the moment of expiration has already passed?

mauroservienti:
> I'm thinking that the `DELETE` statement that does the receive should return a diff between `Expires` column and `GETUTCDATE()` (the receive moment) and based on this diff check if the moment of expiration has already passed?

Sounds like a plan :smile:, using that approach you are getting rid of the possible clock drift since all the action that we make that are based on time are based on the broker time. Correct?

MarcinHoppe:
@mauroservienti This is the goal. All decisions are made on broker time only.

johnsimons:
Please ensure that as part of releasing this patch, ServiceControl is also updated to use this new version and released

mauroservienti:
Updated the description to keep track of the fact that SC needs to be released as well.

@MarcinHoppe @Particular/platform-dev-squad this is missing a proper task force and as discussed should be connected to an issue and not a PR on its own.

src/NServiceBus.SqlServer/ExpiredMessagesPurger.cs
```diff
@@ -0,0 +1,120 @@
...
+                return;
+            }
+
+            var lockTaken = false;
+            try
+            {
+                Monitor.TryEnter(lockObject, ref lockTaken);
+                if (!lockTaken)
+                {
+                    Logger.DebugFormat("An expired message purge task for table {0} is already running. Nothing to do.", queue);
```
johnsimons: @MarcinHoppe you possible can avoid all the locking if instead of using the Timer to execute every time, just change it to execute once and then reschedule it to execute again, this way you stop possible overlapping and you take into account the time that the callback takes.
So in the callback code, at the end of the method, you call `timer.Change(purgeTaskDelay, Timeout.InfiniteTimeSpan)`

MarcinHoppe: @johnsimons This approach would have an added benefit of distributing all purge tasks more evenly. Current approach will hammer the database every 5 minutes. If there are many queue tables in the database and each purge task takes a different amount of time, the next run of the task will most likely happen at a different time than for the other queue tables.

I will implement the change.

MarcinHoppe: Fixed.

MarcinHoppe:
@Particular/transport-maintainers @ramonsmits Please review.

MarcinHoppe:
I'll take the silence as a good sign and will proceed to release this patch later this week.

@Particular/transport-maintainers @ramonsmits Speak now or forever hold your peace :).

mauroservienti:
Oops, sorry. I reviewed this but I forgot to comment, look good to me. @ramonsmits comments?

MarcinHoppe:
Last comment by @ramonsmits has been addressed (expired messages are purged in order).

adamralph:
> as discussed should be connected to an issue and not a PR on its own.

I agree. I think this is in the domain of responsibility of @Particular/transport-maintainers - they can form and execute policy on this without having to involve the @Particular/platform-dev-squad

mauroservienti:
@MarcinHoppe @tmasternak Is this released?

if yes did we ping the @Particular/servicecontrol-maintainers to ask for a ServiceControl release to include this?

mauroservienti:
@justabitofcode FYI, this should fix the issue we were talking about

MarcinHoppe:
@mauroservienti 2 x Yes. See:

https://github.com/Particular/NServiceBus.SqlServer/releases/tag/2.2.2

and here:

https://github.com/Particular/ServiceControl/issues/689

mauroservienti:
Wonderful!

Issue #155, opened 2016-02-08
Multi-schema support in V3 must support square brackets in the schema name


mauroservienti:
taskforce: @kbaley @tmasternak @weralabaj 

## Problem
In V3 we moved the schema name from the connection string to the endpoint address in the form of `my.endpoint@schema`. We want to be prepared to support specifying not only schema but also database if we wish to provide multi-database support in the future. In other words we want users to be able to specify address in form `my.endpoint@[database].[schema]`. 
To be able to support such format it would be preferable to have canonical (used internally and on the wire) representation of address in `SqlServer` transport in form `my.endpoint@[database].[schema]` and `my.endpoint@[schema]`.

## Why `Core V6`
If we leave the canonical transport address as it is right now we would be not wire compatible between v 3.0 and v 3.X with multi-database support. If v 3.X would sent ReplyTo header in form `my.endpoint@[schema]` v 3.0 could not properly parse it. What it means is that we need add the `[]` in v 3.0 to enable proper multi-database implementation later.

## In Scope of this issue
General rule:
 * Whenever users can specify schema (code or message mapping in config file) we support `[]` and bracket-less syntax. For example the two following samples would have the same result

  > meesageType => "Endpoint@dbo"
  > meesageType => "Endpoint@[dbo]"
  
  > configuration.UseTransport<SqlServer>().DefaultSchema("dbo");
  > configuration.UseTransport<SqlServer>().DefaultSchema("[dbo]");

 * Whenever we pass the transport address to use code we use canonical name with `[]`. For example 

   > configuration.UserTransport<SqlServer>().DefaultSchema("dbo");
   > configuration.UseTransport<SqlServer>().EnableLagacyMultiInstanceMode(ta => ..) //Here ta = "endpoint@[dbo]

## Not in scope for this issue
When we provide multi-database support we should extend the rules for new methods if we add them. For example we might have `DefaultDatabase` method which should support `[]` and `[]`-less as well

## Plan of Attack
 - [x] Implement bracket support in schema
 - [x] Update docs => docs will be updated as part of https://github.com/Particular/NServiceBus.SqlServer/issues/139 by @weralabaj
 - [x] Update ProductionTests

/cc @tmasternak @ramonsmits 
Inception: https://github.com/Particular/CustomerSuccess/issues/166

weralabaj:
@mauroservienti does it mean we should support the square brackets version ONLY?

tmasternak:
[...]

PR #175, opened 2016-03-11
Escape schema name in queue address

kbaley:
Connects to #155 

This ensures the schema name in the queue address is escaped to help distinguish it from other components (e.g. catalog) when we add them in the future.

### DIFF
--- a/src/NServiceBus.SqlServer.UnitTests/QueueAddressTests.cs
+++ b/src/NServiceBus.SqlServer.UnitTests/QueueAddressTests.cs
@@ -23,19 +23,16 @@
 
 
         [Test]
-        [TestCase("table", "schema", "schema")]
-        [TestCase("my.table", "my.schema", "my.schema")]
-        [TestCase("table", "", null)]
-        [TestCase("table", null, null)]
-        [TestCase("table", "[my.schema]", "my.schema")]
-        [TestCase("[my.table]", "[my.schema]", "my.schema")]
-        public void Should_parse_address(string tableName, string schemaName, string expectedSchema)
+        [TestCase("table@schema", "table", "schema")]
+        [TestCase("my.table@my.schema@my.schema", "my.table", "my.schema")]
+        [TestCase("table", "table", null)]
+        [TestCase("table@[my.schema]", "table", "my.schema")]
+        [TestCase("[my.table]@[my.schema]", "[my.table]", "my.schema")]
+        public void Should_parse_address(string transportAddress, string expectedTableName, string expectedSchema)
         {
-            var address = new QueueAddress(tableName, schemaName);
+            var parsedAddress = QueueAddress.Parse(transportAddress);
 
-            var parsedAddress = QueueAddress.Parse(address.ToString());
-
-            Assert.AreEqual(tableName, parsedAddress.TableName);
+            Assert.AreEqual(expectedTableName, parsedAddress.TableName);
             Assert.AreEqual(expectedSchema, parsedAddress.SchemaName);
         }
     }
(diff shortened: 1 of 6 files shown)


src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -20,6 +20,10 @@ public static QueueAddress Parse(string address)
                 var parts = address.Split('@');
                 var tableName = parts[0];
                 var schemaName = parts[1];
+                if (schemaName.StartsWith("[") && schemaName.EndsWith("]"))
+                {
+                    schemaName = schemaName.Substring(1, schemaName.Length - 2);
```
MarcinHoppe: Could this be handled by regex?

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -29,7 +33,14 @@ public static QueueAddress Parse(string address)
 
         public override string ToString()
         {
-            return $"{TableName}@{SchemaName}";
+            var escapedSchema = SchemaName ?? "";
+            if (!string.IsNullOrWhiteSpace(escapedSchema) && !(escapedSchema.StartsWith("[") && escapedSchema.EndsWith("]")))
```
tmasternak: Can we normalize schema in constructor. This way in the whole class we could make assumption that the schema does not contain brackets? We could also extract `UnEscape` method that we could use both in constructor in line 23.

src/NServiceBus.SqlServer.UnitTests/QueueAddressTests.cs
```diff
@@ -0,0 +1,42 @@
...
+
+
+        [Test]
+        [TestCase("table", "schema", "schema")]
+        [TestCase("my.table", "my.schema", "my.schema")]
+        [TestCase("table", "", null)]
+        [TestCase("table", null, null)]
+        [TestCase("table", "[my.schema]", "my.schema")]
+        [TestCase("[my.table]", "[my.schema]", "my.schema")]
+        public void Should_parse_address(string tableName, string schemaName, string expectedSchema)
```
tmasternak: I'm wondering if passing whole transport address string and removing `QueueAddress` constructor would not be better here. Right now this test checks that `QueueAddress` is able to parse what it produces but in practice the address will come from the message mappings. 
In other words what I'm saying is let change signature of `Shoudl_parse_address` to (string transportAddress, string expectedSchema).

tmasternak:
I've done my review.

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
kbaley: That brings up an interesting point. What if the schema actually _is_ something like [my.schema] with the brackets? As it stands, the `Parse` method would not work.

We don't actually support this now (it creates oddly named queue tables) so my feeling is that we could get away with doing as @tmasternak suggests and normalize in the constructor. I'll make that refactoring

src/NServiceBus.SqlServer.UnitTests/QueueAddressTests.cs
kbaley: Thanks for the clarification. My intent with this was specifically to test that we could parse what `QueueAddress` produced. Didn't realize that wouldn't happen much in practice. I'll modify this to take in the full transport address.

tmasternak:
@kbaley I've did some small changes. Mainly formatting and naming formatting.

kbaley:
Thanks @tmasternak. Saw your changes and will make note of that for future reference.

PR #184, opened 2016-03-22
Hotfix 2.2.3

### LINKED ISSUE #182: SQL Server transport version 2.2.2 does not support SQL Server 2008

SQL Server transport version 2.2.2 uses [IIF](https://msdn.microsoft.com/en-us/library/hh213574.aspx) logical function in receive and send queries. IIF function is only available in SQL Server 2012 and later, which causes exceptions when endpoint tries to receive messages from the queue in SQL Server 2008.

## Who's affected

Users of SQL Server transport version 2.2.2 that are running SQL Server 2008.

Users of lower versions of the transport are not affected. Users of later versions of SQL Server are not affected.

## Symptoms

The endpoint is unable to send and receive messages from the queue table. When the endpoint tries to receive messages from the queue, it will log the following warnings:

```
An exception occurred when connecting to the configured SQLServer instance
System.Data.SqlClient.SqlException (0x80131904): Incorrect syntax near the keyword 'IS'.
```

### Plan of Attack
- [x] Convert `IIF` to `CASE`
- [x] [Announcement] (https://github.com/Particular/Announcements/pull/83)

MarcinHoppe:
Replace IIF function with CASE statement in TTBR calculations to support SQL Server 2008.

Resolves #182.

### DIFF
--- a/src/NServiceBus.SqlServer/TableBasedQueue.cs
+++ b/src/NServiceBus.SqlServer/TableBasedQueue.cs
@@ -241,13 +241,13 @@ Adding this index will speed up the process of purging expired messages from the
 
         const string SqlSend =
             @"INSERT INTO [{0}].[{1}] ([Id],[CorrelationId],[ReplyToAddress],[Recoverable],[Expires],[Headers],[Body]) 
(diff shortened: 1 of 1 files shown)


MarcinHoppe:
@Particular/transport-maintainers Builds is green, tests are passing. Please give this review & merge a priority.

PR #186, opened 2016-03-22
Replace IIF with CASE to support SQL Server 2008

MarcinHoppe:
Solves #182 on develop.


MarcinHoppe:
@Particular/transport-maintainers This is ready for review and merge.

PR #191, opened 2016-03-29
Cleanup

SimonCropp:
* remove some redundant doco
 * fix nuspecs
 * remove redundant keywords
 * update coderules

### DIFF
--- a/packaging/nuget/NServiceBus.SqlServer.nuspec
+++ b/packaging/nuget/NServiceBus.SqlServer.nuspec
@@ -15,7 +15,7 @@
     <releaseNotes></releaseNotes>
     <tags>nservicebus servicebus msmq cqrs publish subscribe</tags>
     <dependencies>
-	  <dependency id="NServiceBus" version="[5.0.0-unstable, 6.0.0)" />
+	  <dependency id="NServiceBus" version="[6.0.0-unstable, 7.0.0)" />
     </dependencies>
   </metadata>
   <files>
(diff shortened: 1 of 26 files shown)


mauroservienti:
looks good, 2 questions:

* why are we adding the `internal` keyword to classes and interfaces given that it is the default?
* why removing the `readonly` keyword? it has a clear intent in the end

SimonCropp:
internal is the default. hence it is redundant code. in the same whay u dont need private to know something is private with no permission keyword. and there is no "instance" keyword to tell u something is not static.

readonly is redundant on private things. it is just a meta reminded to stop you from accidentally changing a field you should not. but if you are having trouble remembering the scope of what fields should and shouldn't change the class is too complex and should be broken up. many people confuse readonly with the value proposition of immutability, but clearly they cont understand this when they mark mutable data structures (eg lists and dictionaries) with readonly.

mauroservienti:
> internal is the default. hence it is redundant code. in the same whay u dont need private to know something is private with no permission keyword. and there is no "instance" keyword to tell u something is not static.

OK, so why are we adding back the `internal` keyword, e.g. here: https://github.com/Particular/NServiceBus.SqlServer/pull/191/files#diff-407c93b3321247ad8b3321529f7a164cR7

SimonCropp:
let me fix that

mauroservienti:
looks good, @tmasternak any additional comment?

tmasternak:
> but if you are having trouble remembering the scope of what fields should and shouldn't change the class is too complex and should be broken up

@SimonCropp I don't think that is a valid argument. It's as if you were saying: "You don't need safety belts. If you need them it means you are driving too fast.". 
That being said I like the value of being consistent across repositories and we can have that discussion elsewhere.

PR #195, opened 2016-04-07
update to beta 1 of the core

SimonCropp:
connects to https://github.com/Particular/PlatformDevelopment/issues/706


weralabaj:
One test is failing, we need to investigate it.

tmasternak:
Looks like NHiberante version problem https://builds.particular.net/viewLog.html?buildId=156716&tab=buildResultsDiv&buildTypeId=NServiceBusSqlServer_2RunAcceptanceTests. 

I think we need to wait for updated NHibernate package and update the reference.

Issue #181, opened 2016-03-17
Receiving duplicate events from Version 3 publisher after upgrading subscriber from Version 2 to Version 3


tmasternak:
Taskforce: @SzymonPobiega @danielmarbach @johnsimons 

Let's assume we have Endpoint A and Endpoint B both running V2 of SqlServer. Let's assume further that A is subscribing to events published by B.

In situation when Endpoint A gets updated to V3 of SqlServer transport it will start receiving duplicated messages. The root cause of this behavior is duplicate entry in messages subscription storage that will be created at Endpoint B. Before ugrade B will have ` ["EndpointA", "EventTypeX"]`  entry in subscription storage. After subscribe message from A gets processed a second entry will be stored in form `["EndpointA@schemaA", "EventTypeX"]`.

It's important to note the problem should be solved also for situation when publisher is scaled-out and instances are updated to V3 not all at once. 

The reason for this is that:
 * In V3 the address sent in subscription messages has changed from `tableName` to `tableName@schema`
 * Subscription store holds unique constraint on endpoint address (it does not interpret it in any way) and will treat `EndpointA` and `EndpointA@scheamA` as two different addresses - this btw makes a lot of sense
 * V2 endpoint already has a subscription from when Endpoint A was still running on V2

# Who's affected
All using SqlServer V2 and publish-subscribe that want to migrate to V3. 

# Impact
Because this is a blocker for V3 upgrade it is Large.

# Potential solutions

## Provide legacy subscription mode in V3

Make the `@[schema]` suffix an opt-in feature of V3. V3 would default to V2 behavior of not sending the schema name in subscribe messages. Users would be able to use the new mode (which fixes the coupling problems mentioned in #117) only if the V3 endpoint do not subscribe to any V2 endpoints.

### Problems

 * It is hard to enforce "no V2 subscriptions" rule because subscriptions can be added later (after the upgrade) and there is no way we can detect the fact that such a subscription has been added.
 * We're taking one of the new features away from majority of users who will have mixed V2 and V3 deployments until they can upgrade everything to V3.
 
## Send the schema information in a special header

Before dispatching the subscribe message, remove the schema part from the address and move it to a custom header. The V3 publisher which gets that message would be able to read that header and reconstitute the proper `endpoint@[schema]` address before passing it to the subscription infrastructure. The V2 publisher would just ignore the header.

### Problems

 * The wire format is suboptimal and we are postponing the problem for another release instead of solving it
[...]

PR #198, opened 2016-04-13
MessageDispatcher sorts batch per dispatch consistency and reuses the connections more efficiently

danielmarbach:
Connects to https://github.com/Particular/EndToEnd/issues/3
Relates to #181 

In order to understand the implications of #181 I needed to dig deeper into the Sql Transport code. While doing this I realized we are using unnecessary connections inside the same batch. This PR reuses the existing code but does the following:

It creates two paths. One for `Isolated` dispatch consistency and one for `default` consistency.

Side note: The core has a small design flaw. We are marking all operations in a batch as `Isolated` when the outbox is used. Under the outbox we would actually not required any transaction scope. The only thing we would need to make sure is that we are not participating in the receive transaction. But since the core API doesn't allow to differentiate between immediate dispatch and "outbox" mode we need to treat both consistently.

All operations which are isolated will be send inside the same transaction scope with `RequiresNew` on one newly created connection.

All operations which have default consistency will be sent either by using the connection and transaction provided by the message pump or by a newly created connection and transaction.

So in all the modes when we do not have a connection and transaction provided by the message pump we will essentially have one connection instead N connections while N being the number of messages in the batch.

@Particular/transport-maintainers please review

### DIFF
--- a/src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
+++ b/src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
@@ -1,6 +1,8 @@
 ﻿namespace NServiceBus.Transports.SQLServer
 {
+    using System.Collections.Generic;
     using System.Data.SqlClient;
+    using System.Linq;
     using System.Threading.Tasks;
     using System.Transactions;
     using Extensibility;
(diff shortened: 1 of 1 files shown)


src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
...
-            var queue = new TableBasedQueue(destination);
-
-            //Dispatch in separate transaction even if transaction scope already exists
-            if (operation.RequiredDispatchConsistency == DispatchConsistency.Isolated)
-            {
-                await DispatchInIsolatedTransactionScope(queue, operation).ConfigureAwait(false);
-
-                return;
-            }
+            var defaultConsistencyOperations = operations.UnicastTransportOperations.Where(o => o.RequiredDispatchConsistency == DispatchConsistency.Default).ToList();
```
tmasternak: Can we move this to upper method?

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
         // We need to check if we can support cancellation in here as well?
         public async Task Dispatch(TransportOperations operations, ContextBag context)
         {
-            foreach (var operation in operations.UnicastTransportOperations)
-            {
-                await DispatchUnicastOperation(context, operation).ConfigureAwait(false);
-            }
+            await DispatchIsolatedConsistencyTransportOperations(operations).ConfigureAwait(false);
```
tmasternak: Can we change name to something simpler/shorter like `DispatchAsIsolated` - assuming that we do https://github.com/Particular/NServiceBus.SqlServer/pull/198/files#r59613507

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
...
+
+            await DispatchDefaultConsistencyTransportOperations(operations, context).ConfigureAwait(false);
```
tmasternak: Same as above maybe `DispatchAsNonIsolated`?

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
...
+                var queue = new TableBasedQueue(destination);
+
+                await queue.SendMessage(operation.Message, sqlTransportConnection, sqlTransportTransaction).ConfigureAwait(false);
+            }
+        }
+
+        private async Task DispatchOperationsWithNewConnectionAndTransaction(List<UnicastTransportOperation> defaultConsistencyOperations)
         {
             using (var connection = await connectionFactory.OpenNewConnection().ConfigureAwait(false))
             {
```
tmasternak: '{' is not needed here.

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
...
                 using (var transaction = connection.BeginTransaction())
                 {
-                    await queue.SendMessage(operation.Message, connection, transaction).ConfigureAwait(false);
+                    foreach (var operation in defaultConsistencyOperations)
+                    {
+                        var destination = addressParser.Parse(operation.Destination);
+                        var queue = new TableBasedQueue(destination);
 
+                        await queue.SendMessage(operation.Message, connection, transaction).ConfigureAwait(false);
+                    }
```
tmasternak: New line after `}`

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
...
                     transaction.Commit();
                 }
             }
         }
 
-        async Task DispatchInCurrentTransportTransaction(TransportTransaction transportTransaction, TableBasedQueue queue, UnicastTransportOperation operation)
+        private async Task DispatchIsolatedConsistencyTransportOperations(TransportOperations operations)
```
tmasternak: No need for `private`

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
...
+            foreach (var operation in defaultConsistencyOperations)
+            {
+                var destination = addressParser.Parse(operation.Destination);
+                var queue = new TableBasedQueue(destination);
+
+                await queue.SendMessage(operation.Message, sqlTransportConnection, sqlTransportTransaction).ConfigureAwait(false);
+            }
+        }
+
+        private async Task DispatchOperationsWithNewConnectionAndTransaction(List<UnicastTransportOperation> defaultConsistencyOperations)
```
tmasternak: No need for `private`

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
danielmarbach: And pass into the method? Could do, my taste was on the upper level the properly named high level entry points and on the level of usage the where filtering. When your approach pleases more the overall approach here, then happy to change

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
...
+                await DispatchOperationsWithNewConnectionAndTransaction(defaultConsistencyOperations).ConfigureAwait(false);
                 return;
             }
 
-            await DispatchInCurrentTransportTransaction(transportTransaction, queue, operation).ConfigureAwait(false);
+            await DispatchReusingTheReceiveConnectionAndTransaction(transportTransaction, defaultConsistencyOperations).ConfigureAwait(false);
         }
 
-        async Task DispatchAsSeparateSendOperation(TableBasedQueue queue, UnicastTransportOperation operation)
+        private async Task DispatchReusingTheReceiveConnectionAndTransaction(TransportTransaction transportTransaction, List<UnicastTransportOperation> defaultConsistencyOperations)
```
tmasternak: No need for `private`

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -16,83 +18,92 @@ public MessageDispatcher(SqlConnectionFactory connectionFactory, QueueAddressPar
...
             if (transportTransactionExists == false || InReceiveOnlyTransportTransactionMode(transportTransaction))
             {
-                await DispatchAsSeparateSendOperation(queue, operation).ConfigureAwait(false);
-
+                await DispatchOperationsWithNewConnectionAndTransaction(defaultConsistencyOperations).ConfigureAwait(false);
                 return;
             }
 
-            await DispatchInCurrentTransportTransaction(transportTransaction, queue, operation).ConfigureAwait(false);
+            await DispatchReusingTheReceiveConnectionAndTransaction(transportTransaction, defaultConsistencyOperations).ConfigureAwait(false);
```
tmasternak: Maybe `DispatchUsingReceiveTransaction` would be better name?

tmasternak:
@danielmarbach done with review

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
danielmarbach: Then this repos cleanup is out of date. Ran Ctr+E+F

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
danielmarbach: See above

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
tmasternak: I like the idea of high level entry points my only argument would be that it's not obvious what happens to the collection passed into it. Now when I look at the name `DispatchDefaultConsistencyTransportOperations` it does suggests that the filtering is done - so I'm good with keeping the filtering inside the methods.

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
tmasternak: Yes something is off. I'm looking into it.

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
SzymonPobiega: how about
```
var default =  operations.UnicastTransportOperations.Where(o => o.RequiredDispatchConsistency == DispatchConsistency.Default);
var isolated =  operations.UnicastTransportOperations.Where(o => o.RequiredDispatchConsistency == DispatchConsistency.Default);
DispatchIsolated(isolated);
DispatchNonIsolated(default);
```

I mean, when passing the same collection into two methods, from the caller perspective it is not obvious if both methods wont take the same operations etc. With that code we are explicitly categorizing the operations before passing them to proper dispatch methods.

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
SzymonPobiega: LoL, I've just realized I wrote the same comment above ;-)

danielmarbach:
@tmasternak updated

tmasternak:
@danielmarbach Looks good to me. Let's wait for green build and I'll merge.

danielmarbach:
Green

tmasternak:
@danielmarbach good stuff!

SzymonPobiega:
Awesome teamwork

PR #207, opened 2016-04-21
Break out of the loop when nothing to be received

danielmarbach:

Currently the transport wastes a lot of empty receives in competing consumer scenarios or if in the future we will be doing parallel receives. Here is a scenario to highlight the issue

## TODO

* [x] Can we test this at least with a unit test?

We decided to treat the attempt to test outside this PR

### Steps to repro

* Send in 100k messages into the receiver queue
* Start receiving with concurrency of 100
* Truncate the receiver table while receiving (basically simulating someone stealing messages away)

### Effect

Since the first count will receive 100k messages the for loop in the pump will initialize the upper bound to 100 k and start looping. Eventhough the queue is already empty it will still continue to issues the remaining requests against the server. This can get extremely wasteful especially under higher latency scenarios.

## Changes introduced

* Break out of the loop when nothing to be received
* Responsibility of managing the receive cancellation token is moved into receive strategies

## Before

Here is the effect with SQL profiler.

Truncate issued at 5:30:28 PM

image

It keeps on trying to receive ten thousands of messages until 5:30:38

image

It is not visible on the screen shots but we are speaking of ten thousands of unnecessary peeks from the queue.

## After

As seen on this profiler screen shot after the truncate we only try ten peeks and the we realize that the queue is empty and jump out. The number ten is totally arbitrary. It is perfectly possible that we would see up to the number of concurrency try peeks. But definitely not more. After that it jumps back into the polling mode.

image


danielmarbach:
@Particular/sqlserver-transport-maintainers ping

danielmarbach:
// cc @tmasternak @SzymonPobiega

ramonsmits:
Does the receiver know the concept of competing consumer, how many instances are running?

If there are 10 instances, 100 messsages, concurrency limit of 100, it would not be very interesting to try and fetch 100 items and fetching just 10 is probably enough. Taking the message count divided by the number of instances would probably be a good starting point.

Why is the fetch size equal to the concurrency limit? Shouldn't the fetch size be dynamic? Then if all fetches are returning a message this can increase and if it percentage fails the fetch size can decrease by this percentage.

That wouldn't require any large fetch loop cancellation.

danielmarbach:
We wanted to leave each endpoint unaware of the other instances. Any
dynamic approach as we had before gets unnecessary complex. That's why we
ruled it out. The cancellation detection is pretty instant. You'll probably
see in realizy only a few unsuccesful receives. Which is totally ok

src/NServiceBus.SqlServer/Legacy/MultiInstance/LegacyReceiveWithTransactionScope.cs
```diff
@@ -14,8 +14,9 @@ public LegacyReceiveWithTransactionScope(TransactionOptions transactionOptions,
             this.connectionFactory = connectionFactory;
         }
 
-        public async Task ReceiveMessage(TableBasedQueue inputQueue, TableBasedQueue errorQueue, CancellationTokenSource cancellationTokenSource, Func<PushContext, Task> onMessage)
+        public async Task ReceiveMessage(TableBasedQueue inputQueue, TableBasedQueue errorQueue, CancellationTokenSource loopCancellationTokenSource, Func<PushContext, Task> onMessage)
```
tmasternak: Can we change the name of cancellation token? It feaks `MessagePump` details to receiver.

src/NServiceBus.SqlServer/Receiving/MessagePump.cs
```diff
@@ -113,11 +113,18 @@ public void Start(PushRuntimeSettings limitations)
                     return;
                 }
 
+                var loopCancellationTokenSource = new CancellationTokenSource();
+                var loopCancellationToken = loopCancellationTokenSource.Token;
```
tmasternak: New line missing.
Do we need a separate variable? Maybe `loopCancellationToeknSoruce.Token` is good enough?

tmasternak:
Minor points. Otherwise looks good to me :+1:

src/NServiceBus.SqlServer/Receiving/MessagePump.cs
```diff
@@ -188,6 +186,7 @@ await receiveStrategy.ReceiveMessage(inputQueue, errorQueue, tokenSource, pipeli
         Func<PushContext, Task> pipeline;
         Func<TransportTransactionMode, ReceiveStrategy> receiveStrategyFactory;
         IPurgeQueues queuePurger;
+        readonly Func<QueueAddress, TableBasedQueue> queueFactory;
```
SzymonPobiega: Redundant `readonly`

SzymonPobiega:
@tmasternak I'll take it forward.

Issue #206, opened 2016-04-20
Re-consider separating legacy dispatch code


SzymonPobiega:
When implementing https://github.com/Particular/NServiceBus.SqlServer/issues/181 we failed to take the legacy dispatcher into account. Should we try to find some other way to separate the legacy-compliance concern without having a separate class?

tmasternak:
Let me give the arguments for having `Legacy` classes (not only Dispatcher) in the first place the have a proper context:
 * LegacyMultiInstanceMode is legacy :P and moving code responsible for implementing it should make it easy to remove them in v4 of SqlServer transport
 * The key distinction between any Legacy and NonLegacy counterpart classes, is that for Legacy-type logic, transport address may get mapped to a distinct, potentially new SqlConnection (different sql instance). This enables encapsulation of complexity when setting up transactions and/or connections for doing any sql level operations
 * Legacy mode supports only `TransactionScope` receive strategy and e.g. `LegacyDispatcher` could be simplified making such assumptions about that.

tmasternak:
Having all above in mind my first observation would be that in current design we failed to identify parts that will change for the same reasons. In this particular case the way we dispatch messages both for `Legacy` and `NonLegacy` mode.
It might be so that we should use composition to extract the common (read: changing for the same reasons) parts.

SzymonPobiega:
PR opened #209 . Closing.

PR #209, opened 2016-04-21
Removed LegacyMessageDispatcher and fixed defects in LegacyTableBasedQueueDispatcher

SzymonPobiega:

Implements suggestions from #206 and fixes a bug found in LegacyTableBasedQueueDispatcher related to non-atomic publishing of events to multiple subscribers.


SzymonPobiega:
@tmasternak @Particular/sqlserver-transport-maintainers please review.

src/NServiceBus.SqlServer/Sending/IQueueDispatcher.cs
```diff
@@ -0,0 +1,12 @@
+namespace NServiceBus.Transports.SQLServer
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Extensibility;
+
+    interface IQueueDispatcher
+    {
+        Task DispatchAsNonIsolated(List<MessageWithAddress> defaultConsistencyOperations, ContextBag context);
+        Task DispatchAsIsolated(List<MessageWithAddress> isolatedConsistencyOperations, ContextBag context);
```
tmasternak: Let's remove the context from `DispatchAsIsolated`. It's not used in any implementation

src/NServiceBus.SqlServer/Legacy/MultiInstance/LegacyTableBasedQueueDispatcher.cs
```diff
@@ -3,41 +3,47 @@ namespace NServiceBus.Transports.SQLServer.Legacy.MultiInstance
...
                 {
-                    await queue.SendMessage(operation.Message, connection, null).ConfigureAwait(false);
+                    var queue = new TableBasedQueue(operation.Address);
+
+                    using (var connection = await connectionFactory.OpenNewConnection(queue.TransportAddress).ConfigureAwait(false))
+                    {
+                        await queue.SendMessage(operation.Message, connection, null).ConfigureAwait(false);
+                    }
                 }
+                scope.Complete();
```
tmasternak: New line missing

src/NServiceBus.SqlServer/Legacy/MultiInstance/LegacyTableBasedQueueDispatcher.cs
```diff
@@ -3,41 +3,47 @@ namespace NServiceBus.Transports.SQLServer.Legacy.MultiInstance
...
-                    await queue.SendMessage(operation.Message, connection, null).ConfigureAwait(false);
+                    var queue = new TableBasedQueue(operation.Address);
 
-                    scope.Complete();
+                    using (var connection = await connectionFactory.OpenNewConnection(queue.TransportAddress).ConfigureAwait(false))
+                    {
+                        await queue.SendMessage(operation.Message, connection, null).ConfigureAwait(false);
+                    }
                 }
+                scope.Complete();
```
tmasternak: New line missing

src/NServiceBus.SqlServer/Sending/IQueueDispatcher.cs
```diff
@@ -0,0 +1,12 @@
+namespace NServiceBus.Transports.SQLServer
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Extensibility;
+
+    interface IQueueDispatcher
+    {
+        Task DispatchAsNonIsolated(List<MessageWithAddress> defaultConsistencyOperations, ContextBag context);
```
tmasternak: In both signatures we can have just `operations` as argument name

tmasternak:
@SzymonPobiega I've added a commit with review changes. The biggest one was to remove `context` from one of `Dispatcher`'s methods. That resulted in changes that had to be done in test.

Do we want to move those tests to `Integration`? Currently they are in `Unit` tests project which seems wrong to me.

danielmarbach:
@SzymonPobiega I agree with @tmasternak the tests should go into the integration proj.

src/NServiceBus.SqlServer/Legacy/MultiInstance/LegacyTableBasedQueueDispatcher.cs
```diff
@@ -3,41 +3,47 @@ namespace NServiceBus.Transports.SQLServer.Legacy.MultiInstance
...
-                var queue = new TableBasedQueue(operation.Address);
-
-                //If dispatch is not isolated then transaction scope has already been created by <see cref="LegacyReceiveWithTransactionScope"/>
-                using (var connection = await connectionFactory.OpenNewConnection(queue.TransportAddress).ConfigureAwait(false))
+                foreach (var operation in operations)
                 {
-                    await queue.SendMessage(operation.Message, connection, null).ConfigureAwait(false);
+                    var queue = new TableBasedQueue(operation.Address);
+
+                    using (var connection = await connectionFactory.OpenNewConnection(queue.TransportAddress).ConfigureAwait(false))
```
SimonCropp: so since each of these are operating on their own connection does it make sense to do a task.whenall?

tmasternak: I think it would enable us opening more connections in parallel which is good but we should do it only if we do not sacrifice readability/maintainability.

tmasternak:
I'm moving the test to `Integration` project.

tmasternak:
The test have been moved. I've also changed the level at which we are testing from `TableBasedQueueDispatcher` to `MessageDispatcher` this way the tests do not depend on the fact that for `IsolatedDispatch` consistency the dispatch logic is the same no matter if send is done from handler or not. /cc: @SzymonPobiega

danielmarbach:
Approval tests failing

PR #211, opened 2016-04-22
Read small values async and stream large values

danielmarbach:

* Reduces the number of boxing operations in the read path
* Uses sequential access and streaming to asynchronously get data from the server.

Tested it together with @SzymonPobiega with payloads up to 300 KB

## Readings
http://blogs.msdn.com/b/adonet/archive/2012/04/20/using-sqldatareader-s-new-async-methods-in-net-4-5-beta.aspx
https://support.microsoft.com/EN-US/kb/312855

@Particular/sqlserver-transport-maintainers please review


src/NServiceBus.SqlServer/Queuing/MessageParser.cs
```diff
@@ -34,14 +41,34 @@ internal static Message ParseRawData(object[] rowData)
...
+            }
+        }
+
+        static async Task<MemoryStream> GetBody(SqlDataReader dataReader, int bodyIndex)
+        {
+            var memoryStream = new MemoryStream();
+            // Null values will be returned as an empty (zero bytes) Stream.
+            using (var stream = dataReader.GetStream(bodyIndex))
+            {
+                await stream.CopyToAsync(memoryStream).ConfigureAwait(false);
```
SimonCropp: why do we need to copy to a memory stream? cant we just return the underlying stream?

src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
```diff
@@ -98,26 +68,16 @@ public Task SendMessage(OutgoingMessage message, SqlConnection connection, SqlTr
...
             {
                 // ReSharper disable once MethodSupportsCancellation
-                // ExecuteReaderAsync throws InvalidOperationException instead of TaskCancelledException with localized exception message 
-                using (var dataReader = await command.ExecuteReaderAsync(CommandBehavior.SingleRow).ConfigureAwait(false))
-                {
-                    if (await dataReader.ReadAsync(token).ConfigureAwait(false))
-                    {
-                        var rowData = new object[1];
-                        dataReader.GetValues(rowData);
+                var numberOfMessages = (int) await command.ExecuteScalarAsync().ConfigureAwait(false);
```
SimonCropp: why would we not pass the CancellationToken to ExecuteScalarAsync?

src/NServiceBus.SqlServer/Queuing/MessageParser.cs
```diff
@@ -96,13 +123,13 @@ internal static object[] CreateRawMessageData(OutgoingMessage message)
             return data;
         }
 
-        static T GetNullableValue<T>(object value)
+        static async Task<T> GetNullableValueAsync<T>(SqlDataReader dataReader, int index)
         {
-            if (value == DBNull.Value)
+            if (!await dataReader.IsDBNullAsync(index).ConfigureAwait(false))
```
SimonCropp: minor but i would negate this to remove the `!` and make it a little simpler to read

src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
danielmarbach: Happy to do this in another PR. Currently we don't do it consistently that
is correct

src/NServiceBus.SqlServer/Queuing/MessageParser.cs
danielmarbach: We would need to change the way we deal with connections (currently we tear
down the resder and connection) and that stream would not be rewindable.
Also another PR if we want to try it

src/NServiceBus.SqlServer/Queuing/MessageParser.cs
danielmarbach: Good point

src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
tmasternak: I think we do apply it consistently. There are two place that we don't do it:
 * When waiting for task to finish during shutdown 
 * Place mentioned above. It is due to the fact that this method when cancelled thorws `InvalidOperationException` instead of `TaskCanceledException` which we depend on [here](https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Receiving/MessagePump.cs#L88)

tmasternak: That being said I don't know what is the behavior of `ExecueScalarAsync`.

SzymonPobiega:
After discussing (again) with @tmasternak we decided to not include this in the first beta. While this PR is most likely going to improve the performance, we made this decision because we could not come up with an automated test that proves consistently the benefits. We think it is better to not block beta and take the time between the beta and RC to performance-tune the transport properly. 

The biggest issues (likely not related to this PR) we found was that
 * It is hard to measure the receive performance in isolation. Single-threaded tests based on `TableBasedQueue`, even though quite low-level still result in more than 20% differences between the test runs (10K messages for each run)
 * Mentioned tests show that receiving 10K messages, 1KB each takes roughly the same time as receiving 10K 50KB messages while receiving same number of 5KB messages takes significantly longer. This shows something big is off and until we eliminate that big thing it is hard to performance-tune the smaller bits

SzymonPobiega:
@danielmarbach hope you don't mind postponing it until after the beta and that you will join the TF for the performance tuning. Currently I see three main issues:
 * Make the tests repeatable (might involve some infra work such as restarting the SQL Server process before the test etc.)
 * Investigate alternative header parsing (currently reading the text field and then convert it to byte array so that it can be parsed as string...)
 * Investigate if we can alter the reader management to not copy the body into the memory stream

danielmarbach:
I fundamentally disagree with that statement. Let's face it: the testability was not really given before this PR. It follows the best practices of accessing data reader. I tested it with 100 of thousands messages of different sizes on-prem and in the cloud. I see no reason to delay it

danielmarbach:
Let's include me in a call after 3PM and try to convince me ;)

SzymonPobiega:
@danielmarbach :+1:

SzymonPobiega:
Here are more precise test results (90 percentile values in ticks) with following settings:
 * Local SQL Server instance
 * Single-threaded native ADO.NET transaction receive

### Before the change

```
Receiving 20000 1KB messages took 37747 ms. 1,88735 per message. 8155 @90 percentile
Receiving 20000 2KB messages took 42810 ms. 2,1405 per message. 8551 @90 percentile
Receiving 20000 5KB messages took 54697 ms. 2,73485 per message. 11519 @90 percentile
Receiving 20000 7KB messages took 53889 ms. 2,69445 per message. 11151 @90 percentile
Receiving 20000 10KB messages took 34173 ms. 1,70865 per message. 6267 @90 percentile
Receiving 20000 15KB messages took 35696 ms. 1,7848 per message. 6147 @90 percentile
Receiving 20000 20KB messages took 35670 ms. 1,7835 per message. 6151 @90 percentile
Receiving 20000 30KB messages took 42825 ms. 2,14125 per message. 7275 @90 percentile
Receiving 20000 40KB messages took 44088 ms. 2,2044 per message. 7111 @90 percentile
Receiving 20000 50KB messages took 50478 ms. 2,5239 per message. 8075 @90 percentile
Receiving 20000 100KB messages took 74022 ms. 3,7011 per message. 13471 @90 percentile
Receiving 20000 200KB messages took 151173 ms. 7,55865 per message. 22911 @90 percentile
```

### After the change

```
Receiving 20000 1KB messages took 35809 ms. 1,79045 per message. 6907 @90 percentile
Receiving 20000 2KB messages took 42755 ms. 2,13775 per message. 8807 @90 percentile
Receiving 20000 5KB messages took 53931 ms. 2,69655 per message. 11711 @90 percentile
Receiving 20000 7KB messages took 54953 ms. 2,74765 per message. 11831 @90 percentile
Receiving 20000 10KB messages took 33406 ms. 1,6703 per message. 6167 @90 percentile
Receiving 20000 15KB messages took 34413 ms. 1,72065 per message. 6359 @90 percentile
Receiving 20000 20KB messages took 35810 ms. 1,7905 per message. 6231 @90 percentile
Receiving 20000 30KB messages took 38805 ms. 1,94025 per message. 6603 @90 percentile
Receiving 20000 40KB messages took 42001 ms. 2,10005 per message. 6915 @90 percentile
Receiving 20000 50KB messages took 48864 ms. 2,4432 per message. 7495 @90 percentile
Receiving 20000 100KB messages took 61494 ms. 3,0747 per message. 10415 @90 percentile
Receiving 20000 200KB messages took 90182 ms. 4,5091 per message. 15559 @90 percentile
```

Which clearly shows that the new version has advantage over the old one for messages larger than 50KB. This is probably related to the new `async` way of loading the body. I think it is enough to prove the usefulness of this PR.

SzymonPobiega:
@danielmarbach it seems like there is a problem with `When_a_corrupted_message_is_received` test: the `GetValues` call on the data reader returns empty values when a header is corrupted. I guess the data reader does not like mixing the sequential access with `GetValues`.  If I resolve it, I'll merge the pull.

danielmarbach:
@SzymonPobiega nice find

SzymonPobiega:
@tmasternak @SimonCropp @danielmarbach please review my last commit.

SzymonPobiega:
OK, I've rebased it on top of @SimonCropp 's clean up and could not resist the urge to clean up the `MessageRow` class. I ended up refactoring `MessageRow` and `TableBasedQueue` to push responsibilities to the row class. Seems to be green. Please review.

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
```diff
@@ -0,0 +1,198 @@
...
+    using System.Data;
+    using System.Data.SqlClient;
+    using System.IO;
+    using System.Threading.Tasks;
+    using Logging;
+    using static System.String;
+
+    class MessageRow
+    {
+        MessageRow()
```
SimonCropp: isnt this redundant with the below constructor?

SzymonPobiega: Nope. The `Read` factory method uses it.

tmasternak:
Doing review in separate commit. Should be ready soon.

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
SimonCropp: ahh ok

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            var row = new MessageRow();
+            for (var col = 0; col < columns.Length; col++)
+            {
+                await columns[col].Read(dataReader, row, col).ConfigureAwait(false);
+            }
+            return row;
+        }
+
+        public MessageReadResult TryParse()
```
tmasternak: I think that `TryParse` and `Read` should be merged. They used only togather.

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
```diff
@@ -0,0 +1,198 @@
...
+    using Logging;
+    using static System.String;
+
+    class MessageRow
+    {
+        MessageRow()
+        {
+        }
+
+        public MessageRow(OutgoingMessage message)
```
tmasternak: Can we have something like `public MessageRow Parse(OutgoingMessage message)` instead of constructor and private parameterless constructor?

src/NServiceBus.SqlServer/Queuing/MessageReadResult.cs
```diff
@@ -2,31 +2,31 @@ namespace NServiceBus.Transports.SQLServer
 {
     struct MessageReadResult
     {
-        MessageReadResult(Message message, bool poison, object[] dataRecord)
+        MessageReadResult(Message message, MessageRow poisonMessage, bool poison)
```
danielmarbach: Do we need both the poisonMessage and the bool flag? Probably the answer is yes because it is a struct but I wanted to double check

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
danielmarbach: Why not just pass in body and headers instead of OutgoingMessage?

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// </summary>
+        /// <remarks>
+        /// We are assuming sequential access, order is important
+        /// </remarks>
+        public static async Task<MessageRow> Read(SqlDataReader dataReader)
+        {
+            var row = new MessageRow();
+            for (var col = 0; col < columns.Length; col++)
+            {
+                await columns[col].Read(dataReader, row, col).ConfigureAwait(false);
```
danielmarbach: Is this really improving anything? Instead of being able to watch in the method what it does, we have now a foreach loop, a mapping with two lambda bodies one of which is async, more lambda and inner class allocations, dictionary lookups. Since we optimize for the fast pass the non generic approach is simpler to understand and maintain. Not that we add many columns and have to be able to extend it.

src/NServiceBus.SqlServer/Queuing/MessageReadResult.cs
danielmarbach: We could just make it a `Nullable<MessageRow>`

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
tmasternak: Methods have been merged.

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
tmasternak: Changed constructor method to `From(Dictionary.Of(string, string), byte[])`.

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
tmasternak: I've inlined both readying and preparing command (all logic encapsulated previously in ColumInfo instances)

tmasternak:
@SimonCropp can you have a look? If you are good with this I'll merge.

src/NServiceBus.SqlServer.UnitTests/Queueing/MessageParsingPerformanceTests.cs
```diff
@@ -0,0 +1,146 @@
...
+    using System.Diagnostics;
+    using System.Threading.Tasks;
+    using HdrHistogram;
+    using NUnit.Framework;
+    using Transports;
+    using Transports.SQLServer;
+
+    [TestFixture]
+    [Explicit]
+    public class MessageParsingPerformanceTests
```
SimonCropp: what is the plan for this moving forward? should it be run on each release to see if we have perf degradation?

src/NServiceBus.SqlServer.UnitTests/Queueing/MessageParsingPerformanceTests.cs
```diff
@@ -0,0 +1,146 @@
...
+                        }
+                        transaction.Commit();
+                    }
+                    var ticks = Stopwatch.GetTimestamp() - startTimestamp;
+                    histogram.RecordValue(ticks);
+                }
+                watch.Stop();
+                var perMessage = (double)watch.ElapsedMilliseconds / MessageCount;
+                var at90per = histogram.GetValueAtPercentile(90);
+                Console.WriteLine($"Receiving {MessageCount} {bodySizeInKB}KB messages took {watch.ElapsedMilliseconds} ms. {perMessage} per message. {at90per} @90 percentile");
```
SimonCropp: why console write? shouldnt this be Trace?

SimonCropp:
so the change looks good. however is leaving the perf test in there really valuable? unless we can rerun the same test on multiple versions with the same hardware it seems moving forward they would be of less use. should they perhaps instead be in the EndToEnd repo? 

or am i missing something?

tmasternak:
@SimonCropp I've removed the test. It was used during refactorings to make sure the changes have positive effect in that regard.

SzymonPobiega:
@SimonCropp I think there is not enough value in the test to keep it alive, at least for now.

SimonCropp:
ok i will smoke test this on some samples and then merge

PR #215, opened 2016-04-22
InnerReceive yields to enforce asynchronous continuation to avoid costly synchronous connection pooling and enlistment on pump task.

### LINKED ISSUE #214: Receive Performance with higher latency can be increased with parallelism
I still do not grok this issue entirely. So please bear with me.

I found out that if we receive in parallel (on my machine with 8 threads, more doesn't help) we can get much higher throughput especially in higher-latency scenarios.

```
Sql Azure

Concurrency 100, 1 pump
Initalizing the bus took 0:00:00:01.5281351
Receiving #10000 of msgs over the bus took 0:00:04:14.7216617

Concurrency 100, 8 parallel pump
Initalizing the bus took 0:00:00:03.9082233
Receiving #10000 of msgs over the bus took 0:00:00:41.8102496
```

```
Sql Local

Concurrency 100, 1 pump
Initalizing the bus took 0:00:00:00.7398390
Receiving #100000 of msgs over the bus took 0:00:01:38.7345426

Concurrency 100, 8 parallel pump
Initalizing the bus took 0:00:00:01.1255246
Receiving #100000 of msgs over the bus took 0:00:01:23.8303070
```

I couldn't explain why, that's why I started digging a bit deeper. I first suspected the [queue count peek](https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Receiving/MessagePump.cs#L104). But even hardcoding it in the pump doesn't change a thing.

It seems that `SqlConnection.OpenAsync` is a blocking operation although the name suggests otherwise. With `TransactionScope` in place the DTC enlistment seems to be synchronous. Locally you don't see much of an effect of this behavior. But if you connect for example to Sql Azure this becomes apparent. The `OpenAsync` call eats up the majority of the read performance.

## Local OpenAsync with TxScope

The offender is [`SNIReadSyncOverAsync`](http://referencesource.microsoft.com/#System.Data/System/Data/SqlClient/TdsParserStateObject.cs,1944). Locally 5% of the costs in the call stack

image

## Sql Azure OpenAsync with TxScope

The offender is [`SNIReadSyncOverAsync`](http://referencesource.microsoft.com/#System.Data/System/Data/SqlClient/TdsParserStateObject.cs,1944). Locally 99.7% of the costs in the call stack

image

That would explain why 8 parallel receivers are much faster. So I figured let me just switch the transaction mode to `SendWithAtomicReceive` to get rid of the costly enlistment.

`Receiving #10000 of msgs over the bus took 0:00:03:52.587666`

It improved the overall message consumption time in Azure roughly around 20 seconds. So I can't still explain why 8 parallel receives are so much faster. There must be more operations which are blocking inside the `System.Data` namespace.

You can see in the runs on top in this description that locally having 8 parallel receive is not a big difference, but still there is a difference. The higher the latency gets, the more you'll see the difference in throughput.

Assuming all System.Data call we do asynchronously are actually truly asynchronous it should not matter. Any thoughts?

danielmarbach:

Fixes #214 

This will increase performance of receives in order of magnitude on high latency scenarios but also low latency scenarios. On my machine I saw up to 6 times more throughput with Azure SQL S3.

Deeper analysis have shown that the `SqlConnection` constructor surrounded with a `TransactionScope` will do very costly operations synchronously on the calling thread such as [connection pooling](http://referencesource.microsoft.com/#System.Data/System/Data/ProviderBase/DbConnectionFactory.cs,208) and DTC enlistment.

## Reproduction Sample

Uses Sql.cs and MessageParser.cs from #211 

```
   internal class Program
    {
        private static void Main(string[] args)
        {
            var receives = new ConcurrentDictionary<Task, Task>();
            var stopwatch = Stopwatch.StartNew();
            var semaphore = new SemaphoreSlim(100);
            var runner = Task.Run(async () =>
            {
                for (var i = 0; i < 10000; i++)
                {
                    //Console.Write(i);
                    await semaphore.WaitAsync().ConfigureAwait(false);

#pragma warning disable 4014
                    var receive = Receive().ContinueWith(t =>
                    {
                        Task dummy;
                        receives.TryRemove(t, out dummy);
                        semaphore.Release();
                        Console.Write(".");
                    }, TaskContinuationOptions.ExecuteSynchronously);
                    receives.TryAdd(receive, receive);
#pragma warning restore 4014
                }
            });

            runner.Wait();
            Task.WaitAll(receives.Values.ToArray());
            stopwatch.Stop();

            Console.WriteLine(stopwatch.Elapsed);

            Console.ReadLine();
        }

        static async Task Receive()
        {
            using(var tx = new TransactionScope(TransactionScopeOption.Required, new TransactionOptions { IsolationLevel = IsolationLevel.ReadCommitted }, TransactionScopeAsyncFlowOption.Enabled))
            using (
                var connection = new SqlConnection(@"SQL Connection string")
                )
            {
                await connection.OpenAsync().ConfigureAwait(false);
                using (var command = new SqlCommand(string.Format(Sql.ReceiveText, "dbo", "YourTable")))
                {
                    command.Connection = connection;

                    using (var reader = await command.ExecuteReaderAsync(CommandBehavior.SingleRow | CommandBehavior.SequentialAccess).ConfigureAwait(false))
                    {
                        try
                        {
                            if (await reader.ReadAsync().ConfigureAwait(false))
                            {
                                var message = await MessageParser.ParseRawData(reader).ConfigureAwait(false);
                                Console.Write(message.TransportId.Substring(0,1));
                            }
                        }
                        catch (Exception ex)
                        {
                            Console.WriteLine(ex);
                            Console.Write("!");
                        }
                    }
                }

                tx.Complete();
            }
        }
    }
```

This code takes over roughly 270 seconds on my machine to process 10 k messages.

You can see the evil things that happen in the constructor and how it eats the pump thread in the following screenshots

image

Thread 3 is the pump thread. It pays the majority of the costs. Other threads will execute mostly the continuations.

image

## Solution

Super simple.

Enforce the code before opening the connection to yield asynchronously with [Task.Yield](https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.yield(v=vs.110).aspx).

So the above shown code would look like this:

```
...
        static async Task Receive()
        {
            await Task.Yield();  // MAGIC HAPPENS HERE ;)

            using(var tx = new TransactionScope(TransactionScopeOption.Required, new TransactionOptions { IsolationLevel = IsolationLevel.ReadCommitted }, TransactionScopeAsyncFlowOption.Enabled))
            using (
                var connection = new SqlConnection(@"SQL Connection string")
                )
            {

...
```

For all nitpickers. `Task.Yield` doesn't provide `ConfigureAwait` ;)

Then the code is able to process 10K messages in 38 seconds on my machine.

An the profiler results looks like

Work is evenly distributed (more or less ;) )

image

all involved threads will open up connections

image

### Before

```
Concurrency 100, 10 K Messages
Initalizing the bus took 0:00:00:01.5956504
Receiving #10000 of msgs over the bus took 0:00:04:11.7019003
```

### After

```
Concurrency 100, 10 K Messages
Initalizing the bus took 0:00:00:01.6650525
Receiving #10000 of msgs over the bus took 0:00:00:40.3449894
```

## Readings
 * [When to use Task.Yield](http://stackoverflow.com/questions/22645024/when-would-i-use-task-yield)

### DIFF
--- a/src/NServiceBus.SqlServer/Receiving/MessagePump.cs
+++ b/src/NServiceBus.SqlServer/Receiving/MessagePump.cs
@@ -144,6 +144,10 @@
             {
                 try
                 {
+                    // We need to force the method to continue asynchronously because SqlConnection
+                    // in combination with TransactionScope will apply connection pooling and enlistment synchronous in ctor.
+                    await Task.Yield();
+
                     await receiveStrategy.ReceiveMessage(inputQueue, errorQueue, tokenSource, pipeline)
                         .ConfigureAwait(false);
 


danielmarbach:
@Particular/sqlserver-transport-maintainers please review and merge

Can you guys make sure you also bring in my other PRs and rebase those if necessary since I'm away next week. Thx!

dvdstelt:
Nicely done, @danielmarbach 
A description of Task.Yield() can be found here : http://stackoverflow.com/questions/22645024/when-would-i-use-task-yield/22645114#22645114
(I Googled it ;-) )

danielmarbach:
Thanks. Btw I linked it in the text :)

danielmarbach:
@simoncropp this might be relevant for your native persistence as well

tmasternak:
@danielmarbach Great job!

ramonsmits:
Results I did on the on Beta 8 core & Beta 5 SqlTransport. It proves that in most cases using Task.Yield is faster in most tests.

## Atomic

|  **Concurrency** | **DisableYield** | **Throughput msg/s** |
|  ------: | ------ | ------: |
|  32 |  | 1237.0 |
|  8 | true | 1103.0 |
|  8 |  | 1083.7 |
|  128 |  | 1023.1 |
|  128 | true | 832.3 |
|  32 | true | 793.3 |
|  1 |  | 599.0 |
|  1 | true | 553.1 |

## None

|  **Concurrency** | **DisableYield** | **Throughput msg/s** |
|  ------: | ------ | ------: |
|  32 | true | 1155.6 |
|  8 | true | 1104.7 |
|  8 |  | 1082.5 |
|  32 |  | 1009.4 |
|  128 | true | 996.8 |
|  128 |  | 919.3 |
|  1 | true | 570.6 |
|  1 |  | 550.7 |

## Receive

|  **Concurrency** | **DisableYield** | **Throughput msg/s** |
|  ------: | ------ | ------: |
|  32 |  | 1236.0 |
|  8 |  | 1171.0 |
|  128 |  | 952.8 |
|  8 | true | 940.6 |
|  128 | true | 892.5 |
|  32 | true | 820.2 |
|  1 | true | 582.6 |
|  1 |  | 574.5 |

## Transactional

|  **Concurrency** | **DisableYield** | **Throughput msg/s** |
|  ------: | ------ |------:|
|  128 |  | 1146.8 |
|  8 |  | 1048.7 |
|  32 |  | 1028.4 |
|  8 | true | 1004.0 |
|  32 | true | 854.8 |
|  128 | true | 796.7 |
|  1 |  | 598.8 |
|  1 | true | 563.9 |

PR #220, opened 2016-04-28
WIP: dont route to NServiceBus.SqlServer.CallbackQueue for audit messages

### LINKED ISSUE #202: Audit message is sent to the address from `NServiceBus.SqlServer.CallbackQueue` header
### Symptoms

Audit message is sent to the address from `NServiceBus.SqlServer.CallbackQueue` header.

### Who's affected

All users of version 3 who also have endpoints using version 2.

### Root cause

The legacy callback handling behavior overrides the destination in the routing stage if the *incoming* message is a reply. The intent of the author was probably to override it if the current message being sent is a reply. This stage happens to be executed also for non-user messages like audits and errors. In this case the audit message is redirected to the destination from the legacy callback queue header. Even if the intent was checked on the current message it would not work because it would override the destination of audited or forwarded replies.

### Repro

Run the SQLServer/NHibernate outbox sample with V2 version of `Receiver` and V3 version of `Sender` to notice `Receiver` blowing up with unexpected message (which has been routed to it instead of the audit queue)

### Plan of attack

 * [ ] Evaluate a solution which copies the value from callback queue header to `ReplyTo` header in the message pump
 * [ ] Evaluate a solution which copies the value from callback queue header to `ReplyTo` header in the handling stage (before executing the handlers)
 * [ ] Copy the fix to RabbitMQ transport

SimonCropp:
fixes #202

### DIFF
--- a/src/NServiceBus.SqlServer/Legacy/Callbacks/OverrideOutgoingReplyAddressBehaviorBasedOnLegacyHeader.cs
+++ b/src/NServiceBus.SqlServer/Legacy/Callbacks/OverrideOutgoingReplyAddressBehaviorBasedOnLegacyHeader.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Transports.SQLServer
 {
     using System;
+    using System.Diagnostics;
     using System.Threading.Tasks;
     using Extensibility;
     using Pipeline;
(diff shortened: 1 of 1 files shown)


johnsimons:
The current approach to check if a message is being routed to be audited seems not ideal
```c#
return context.Message.Headers.ContainsKey(Headers.ProcessingEndpoint);
```
Looking at this problem it seems we need some kind of "intent" on the `context` itself, something like Intent.Audit, Intent.Error, ...

PR #221, opened 2016-05-02
Changing ReplyTo with CallbackQueue for legacy callback support

### LINKED ISSUE #202: Audit message is sent to the address from `NServiceBus.SqlServer.CallbackQueue` header

tmasternak:
Fixes #202 


tmasternak:
@SimonCropp this is the alternative approach to fixing legacy callback messages. I've branched if of #211 because changes have been made around message parsing. Pretty much all action happens in [here](https://github.com/Particular/NServiceBus.SqlServer/pull/221/files#diff-b85b5b4ccb8367cf5f56df95bde2e819R97).

Please note that I have removed any conditions on MessageIntent and now any type of message that is received will have `ReplyTo` overridden if `CallbackQueue` header exists.

/cc: @SzymonPobiega

src/NServiceBus.SqlServer/Legacy/Callbacks/LegacyCallbacks.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace NServiceBus.Transports.SQLServer
+{
+    using System.Collections.Generic;
+
+    class LegacyCallbacks
+    {
+        public static void SubstituteReplyToWithCallbackQueueIfExists(Dictionary<string, string> headers)
+        {
+            if (headers.ContainsKey("NServiceBus.SqlServer.CallbackQueue"))
```
SzymonPobiega: Use `TryGetValue`?

tmasternak: :+1:

PR #222, opened 2016-05-04
Fixing flaky att

tmasternak:

/cc: @SzymonPobiega


SzymonPobiega:
It this already fixed in the core (but not yet released)?

tmasternak:
Correct

PR #230, opened 2016-05-11
Opt out from waiting for DONE_IN_PROC messages by setting NO COUNT ON for statements that don't need it

danielmarbach:
Connects to: #231 

Caveat first: This might be an esoteric change.

NO COUNT ON: Stops the message that shows the count of the number of rows affected by a Transact-SQL statement or stored procedure from being returned as part of the result set.

[Official MS statement](https://msdn.microsoft.com/en-us/library/ms189837.aspx)

> SET NOCOUNT ON prevents the sending of DONE_IN_PROC messages to the client for each statement in a stored procedure. For stored procedures that contain several statements that do not return much actual data, or for procedures that contain Transact-SQL loops, setting SET NOCOUNT to ON can provide a significant performance boost, because network traffic is greatly reduced.

Although detailed analysis[ from Dale on his blogs ](http://daleburnett.com/2014/01/everything-ever-wanted-know-set-nocount/)shows that

> The documentation claims that setting NOCOUNT on can have significant network bandwidth savings. As we can see this is on a case-by-case basis determined by the number of non-data returning statements being executed relative to the amount of results returned. A WHILE loop that executes 10,000 times in a procedure that only returns one row will see a great reduction in traffic relative to the size of the entire result set. But a few non-data returning statements in a batch that returns thousands and thousands of rows will see practically zero percentage change in bandwidth.

## WITH NOCOUNT OFF

image

## WITH NOCOUNT ON

image

Costs of the call are non-existing so I don't think it would hurt.

image

Permissions: Requires membership in the public role.

Local analysis shows not big of a difference. We would need to test with higher latency though

https://docs.google.com/a/nservicebus.com/spreadsheets/d/13v7llKhqZgNpU-_7KT76pZvd-ke8TXnigFsu1DbS_qw/edit?usp=sharing

### Sends

image

### Receives

image

## Readings

- https://msdn.microsoft.com/en-us/library/ms189837.aspx
- https://msdn.microsoft.com/en-us/library/dd340553.aspx
- http://daleburnett.com/2014/01/everything-ever-wanted-know-set-nocount/

### DIFF
--- a/src/NServiceBus.SqlServer/Queuing/Sql.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Sql.cs
@@ -5,13 +5,27 @@ namespace NServiceBus.Transport.SQLServer
         internal const string PurgeText = "DELETE FROM {0}.{1}";
 
         internal const string SendText =
-            @"INSERT INTO {0}.{1} ([Id],[CorrelationId],[ReplyToAddress],[Recoverable],[Expires],[Headers],[Body])
-                                    VALUES (@Id,@CorrelationId,@ReplyToAddress,@Recoverable,CASE WHEN @TimeToBeReceivedMs IS NOT NULL THEN DATEADD(ms, @TimeToBeReceivedMs, GETUTCDATE()) END,@Headers,@Body)";
+            @"
+              DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
+              IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON'
+              SET NOCOUNT ON;
 
-        internal const string ReceiveText =
-            @"WITH message AS (SELECT TOP(1) * FROM {0}.{1} WITH (UPDLOCK, READPAST, ROWLOCK) WHERE [Expires] IS NULL OR [Expires] > GETUTCDATE() ORDER BY [RowVersion])
-			DELETE FROM message
-			OUTPUT deleted.Id, deleted.CorrelationId, deleted.ReplyToAddress, deleted.Recoverable, deleted.Headers, deleted.Body;";
+              INSERT INTO {0}.{1} ([Id],[CorrelationId],[ReplyToAddress],[Recoverable],[Expires],[Headers],[Body])
+              VALUES (@Id,@CorrelationId,@ReplyToAddress,@Recoverable,CASE WHEN @TimeToBeReceivedMs IS NOT NULL THEN DATEADD(ms, @TimeToBeReceivedMs, GETUTCDATE()) END,@Headers,@Body);;
+
+              IF(@NOCOUNT = 'ON') SET NOCOUNT ON;
+              IF(@NOCOUNT = 'OFF') SET NOCOUNT OFF;";
+
+        internal const string ReceiveText = @"
+            DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
+            IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON';
+            SET NOCOUNT ON;
+
+            WITH message AS (SELECT TOP(1) * FROM {0}.{1} WITH (UPDLOCK, READPAST, ROWLOCK) WHERE [Expires] IS NULL OR [Expires] > GETUTCDATE() ORDER BY [RowVersion])
+            DELETE FROM message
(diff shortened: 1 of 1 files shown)


danielmarbach:
@Particular/sqlserver-transport-maintainers please review

ramonsmits:
Usage of NO COUNT ON very often has a positive impact when doing multiple queries in a single roundtrip or as the link states doing multiple queries in a single stored procedure.

We are not doing that here hence not noticing a big improvement.

My main concern here is that the 'default' state is not reset.

If the connection was already SET NOCOUNT ON, this PR now sets it to SET NOCOUNT OFF which can have a negative performance impact

I see SET NOCOUNT ON/OFF not as a responsibility of the query but of the connection factory. I rather see it done here:

https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Configuration/SqlConnectionFactory.cs

Please be aware that NHibernate persistence requires NO COUNT OFF in order for its optimistic concurrency to behave correctly.

danielmarbach:
How about this we change it to:

```
SET NOCOUNT ON;
-- INSERT or CTE
DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON';
if(@NOCOUNT = 'ON') SET NOCOUNT ON;
if(@NOCOUNT = 'OFF') SET NOCOUNT OFF;
```

ramonsmits:
I see that `@@OPTIONS` is 2008+, do we support earlier SQL versions? 

I just did a test and @@options only contained the current value if I added a GO:

```sql
SET NOCOUNT ON;

DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON';
if(@NOCOUNT = 'ON') print 'on';
if(@NOCOUNT = 'OFF') print 'off';

SET NOCOUNT OFF;

IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON';
if(@NOCOUNT = 'ON') print 'on';
if(@NOCOUNT = 'OFF') print 'off';
--
on
on
```

vs

```sql
SET NOCOUNT ON;
GO

DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON';
if(@NOCOUNT = 'ON') print 'on';
if(@NOCOUNT = 'OFF') print 'off';

SET NOCOUNT OFF;
GO

DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON';
if(@NOCOUNT = 'ON') print 'on';
if(@NOCOUNT = 'OFF') print 'off';

--
on
off
```

So in order to correctly read @@OPTIONS it seems the SET NOCOUNT ON/OFF should be part of another statement.

ramonsmits:
And as the @@options needs to be read from its state before the INSERT/DELETE your statement would indeed work @danielmarbach

danielmarbach:
2008 and above

danielmarbach:
Implemented NOCOUNT properly so that it is reset to the previously seen value

danielmarbach:
@ramonsmits could there be an impact on the NH persistence ? Since the NOCOUNT is only applied to a single statement I figured not

danielmarbach:
Got a review today from a SQL server MVP. He said specifying NOCOUNT like done in this PR is best practice

ramonsmits:
I would say, lets really measure it and see if it has a benefit. It does add more complexity to the query so if there is no measurable performance difference it should not be merged.

danielmarbach:
I just confirmed this again with another data platform MVP. We should definitely do this.

tmasternak:
@ramonsmits can we use `EndToEnd` tests for that? How complicated would that be?

tmasternak:
Lets reuse AWS instances prepared by @SzymonPobiega and test it and If we seen performance gains let's merge it.

tmasternak:
I'm merging this one. It is a best practice and is scoped only to single statements + we do not share connection with nHiberante persister. If we see problems with performance we can always backtrack.

PR #233, opened 2016-05-13
Peek only non expired messages

danielmarbach:
The current code base does always peek even expired messages, stream them from the server to the client, reads the data row into memory and the checks then if the message is expired. 

The benefit of the previous approach was that expired messages get deleted one by one by the receive loop and if there are leftovers the will be swept away by the purge expired messages cleaner. The drawback is that each and every messages needs to be loaded into the client, and the expired field needs to be checked.

With these changes expired messages will no longer add to the peek count and will no longer be peeked. They will only be swept away when the expired messages cleaner kicks in. The drawback here is that depending on the cleaner interval and the insertion rate the table might fill itself faster. In my opinion this is the better tradeoff then unnecessarily load expired messages to the client and pay the whole price of latency, memory and more.

We might need to think about exposing

PurgeTaskDelayTimeSpan
PurgeBatchSize

to the users. But that is a topic for another discussion or PR.

### DIFF
--- a/src/NServiceBus.SqlServer/Queuing/MessageRow.cs
+++ b/src/NServiceBus.SqlServer/Queuing/MessageRow.cs
@@ -64,9 +64,8 @@
             row.correlationId = await GetNullableAsync<string>(dataReader, 1).ConfigureAwait(false);
             row.replyToAddress = await GetNullableAsync<string>(dataReader, 2).ConfigureAwait(false);
             row.recoverable = await dataReader.GetFieldValueAsync<bool>(3).ConfigureAwait(false);
-            row.timeToBeReceived = await GetNullableValueAsync<int>(dataReader, 4).ConfigureAwait(false);
-            row.headers = await GetHeaders(dataReader, 5).ConfigureAwait(false);
-            row.bodyStream = await GetBody(dataReader, 6).ConfigureAwait(false);
+            row.headers = await GetHeaders(dataReader, 4).ConfigureAwait(false);
+            row.bodyStream = await GetBody(dataReader, 5).ConfigureAwait(false);
 
             return row;
         }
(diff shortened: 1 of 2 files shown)


MarcinHoppe:
@danielmarbach Both purge parameters are exposed via settings, just not documented. This was a deliberate decision on our part. If we merge this PR, then role of these settings will be more important and we probably need to add them as first-class citizens to the code-first API.

Another concern here is how do we select default value of the delay and how do we test the impact of this change on throughput?

MarcinHoppe:
Also, we should probably have an ATT in place that checks that expired messages are indeed purged from the queue table at one point.

Other than that, LGTM.

SzymonPobiega:
I would love to include that in the RC. How can we make sure the change in the query does not slow down receiving if there is no expired messages?

Would that lab environment be helpful here? /cc @tmasternak

MarcinHoppe:
We also need to check what are the performance implications when there is no index on the `Expires` column.

danielmarbach:
I definitely think we should get this in! Can we make some progress on this one?

danielmarbach:
Rebased

MarcinHoppe:
@danielmarbach We should test this thoroughly. Next week? :smile:

tmasternak:
@MarcinHoppe What kind of tests do you have in mind? I'm asking because maybe it's worth to setup something more reusable in AWS?

MarcinHoppe:
@tmasternak Performance tests. I would like to determine if and how this affects performance. Myself and @danielmarbach are planning some perf work on the transport next week.

tmasternak:
Are you going to use `EndToEnd` test or something more specific to `SqlTransport`?

MarcinHoppe:
I think we will first define what we want to measure and then determine how to measure it. Does this qualify as an answer? :grin:

tmasternak:
@SzymonPobiega is working on AWS test environment. I would vote to test it there and merge this one if possible.

tmasternak:
This one is already taking too long to merge. I've done some testing on my local box and execution plans for old and new quires are the same. 
image

If we see any problems with performance once it goes through `EndToEnd` test suite we can backtrack.

PR #252, opened 2016-06-08
Early closing of SQL connection in the ambient transaction mode.

SzymonPobiega:
Closing the SQL connection before invoking the pipe allows other things in the pipe (e.g. persistence) that use the same `TransactionScope` to open their connections without necessarily escalating to DTC (if both connections have the same connection string). The flow would be following:

1. Create `TransactionScope`
1. Open transport connection
1. Receive message
1. Close transport connection
1. Execute the pipe
1. Open persistence connection
1. Handle the message
  1. Enqueue message to be sent (batching)
  1. Open immediate dispatch `TransactionScope`
    1. Open immediate dispatch connection
    1. Send the message
    1. Close the immediate dispatch connection
  1. Dispose immediate dispatch `TransactionScope`
1. Close persistence connection
1. Open batch dispatch connection
1. Send batched messages
1. Close batch dispatch connection
1. Dispose transport `TransactionScope`

NOTE: Because the connection is closed early, there is no way for a handler to get that connection. But this should not be an issue because the handler is free to open its own connection and, if the connection string matches, it would not cause DTC escalation.

[here](https://github.com/Particular/docs.particular.net/pull/1594) a doco pull that assumes this is merged and provides info how to handle handlers that access the transport connection during 2->3 upgrade.

### DIFF
--- a/src/NServiceBus.SqlServer/Receiving/ReceiveWithTransactionScope.cs
+++ b/src/NServiceBus.SqlServer/Receiving/ReceiveWithTransactionScope.cs
@@ -43,8 +43,9 @@
                 using (var pushCancellationTokenSource = new CancellationTokenSource())
                 using (var bodyStream = message.BodyStream)
                 {
+                    connection.Close();
+
                     var transportTransaction = new TransportTransaction();
-                    transportTransaction.Set(connection);
                     transportTransaction.Set(Transaction.Current);
 
                     var pushContext = new PushContext(message.TransportId, message.Headers, bodyStream, transportTransaction, pushCancellationTokenSource, new ContextBag());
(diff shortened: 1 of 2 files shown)


dvdstelt:
@SzymonPobiega : I'm not completely aware what you're trying to achieve here. But from what I understand, if batching is used and the user doesn't open its own connection, then the transaction won't be promoted to DTC. Whereas in the old situation you'd keep the connection open and it always gets promoted to DTC. So the change that it doesn't get promoted is very, very slim, but could happen.

SzymonPobiega:
@dvdstelt in the old (current) code, the batching sends re-use the receive connection so it does not get promoted to DTC. So it seems like the behavior wrt promotion is similar. 

What I am trying to achieve is simplify (and fix) the core related to sharing transaction context between SQL transport and NH persistence. The current one is using a shared connection while what I am proposing here is using separate connections while still sharing a transaction scope.

SzymonPobiega:
@Particular/sqlserver-transport-maintainers do you agree with my reasoning?

SzymonPobiega:
Rebased. @tmasternak plase merge when green.

SzymonPobiega:
Green.

PR #283, opened 2016-07-22
MessageDispatcher no longer uses LINQ and the allocations that come with it

danielmarbach:
LINQ is expensive both in terms of allocs and time it takes. In the core we are trying to avoid it on hot paths.


danielmarbach:
Will add more details soon

tmasternak:
LGTM

PR #290, opened 2016-08-16
[WIP] Overriding queue name via EndpointInstance property

SzymonPobiega:
Connects to https://github.com/Particular/NServiceBus/pull/4064

An acceptance test that verifies the behavior is a part of the parent PR.


MarcinHoppe:
@SzymonPobiega Said this is not necessary because the SQL transport does not support file based routing.

PR #292, opened 2016-08-31
Added settings for queue peeker delay

### LINKED ISSUE #271: add code API for controlling QueuePeeker peek interval
https://github.com/Particular/NServiceBus.SqlServer/blob/3.0.0-Beta2/src/NServiceBus.SqlServer/Receiving/QueuePeeker.cs#L47

propose supported range: 100ms to 10s

HEskandari:
Resolves #271

### DIFF
--- a/src/NServiceBus.SqlServer/Receiving/QueuePeeker.cs
+++ b/src/NServiceBus.SqlServer/Receiving/QueuePeeker.cs
@@ -7,9 +7,10 @@
 
     class QueuePeeker : IPeekMessagesInQueue
     {
-        public QueuePeeker(SqlConnectionFactory connectionFactory)
+        public QueuePeeker(SqlConnectionFactory connectionFactory, QueuePeekerOptions settings)
         {
             this.connectionFactory = connectionFactory;
+            this.settings = settings;
         }
 
         public async Task<int> Peek(TableBasedQueue inputQueue, RepeatedFailuresOverTimeCircuitBreaker circuitBreaker, CancellationToken cancellationToken)
@@ -26,7 +27,7 @@
 
                     if (messageCount == 0)
                     {
-                        await Task.Delay(peekDelay, cancellationToken).ConfigureAwait(false);
+                        await Task.Delay(settings.Delay, cancellationToken).ConfigureAwait(false);
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


HEskandari:
@Particular/sqlserver-transport-maintainers Please review.

src/NServiceBus.SqlServer/Receiving/QueuePeekerOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+            }
+
+            Delay = delay;
+        }
+
+        static void Validate(TimeSpan delay)
+        {
+            if (delay < TimeSpan.FromMilliseconds(100))
+            {
+                var message = "Delay requested is invalid. The value should be greater than 100 ms and less than 10 seconds.";
```
tmasternak: Can we have a single message for both cases?

src/NServiceBus.SqlServer/Receiving/QueuePeeker.cs
```diff
@@ -42,9 +43,9 @@ public QueuePeeker(SqlConnectionFactory connectionFactory)
             return messageCount;
         }
 
-        SqlConnectionFactory connectionFactory;
+        readonly SqlConnectionFactory connectionFactory;
```
SzymonPobiega: redundant `readonly`

tmasternak:
@HEskandari I think we need a doco for the new method as well

tmasternak:
We should be passing the setting also to `LegacyQueuePeeker`

HEskandari:
@tmasternak oh forgot that one completely :) I'll update.

tmasternak:
The current approach will set the same peek interval for all queues which I think is good. (just confirming ;))

src/NServiceBus.SqlServer/Receiving/QueuePeekerOptions.cs
HEskandari: @tmasternak wouldn't that be confusing if the message says the value is "invalid" but still runs? That was why I have used a slightly different wording.

tmasternak: That makes sense @HEskandari. Thanks for clarification.

PR #293, opened 2016-08-31
Added settings for queue peeker delay

HEskandari:
This one replaces #292 and should address all the previous comments.  The documentation will be done as a part of Particular/docs.particular.net#1835


HEskandari:
@Particular/sqlserver-transport-maintainers This should be all ready now. Please review.

Issue #204, opened 2016-04-19
Receiving concurrently more than Default Max Pool Size of messages results in exception.


ramonsmits:
## Reproduction

In V3 of the transport when sending and/or receiving more than 100 (which is default size of SqlConnection pool) messages concurrently the connection pool size reaches it's limit which results in exceptions being thrown.

## PoA

* [x] When connection string is configured via `ConnectionString` during initialization of the transport check if `Max Pool Size` is set or if pooling is disabled and if not we log a WARN and direct to doco.
* [x] When connection is provided with registered callback do a one-time check similar to above and if needed log a WARN

tmasternak:
I think that we need to clearly define a goal when resolving this one. I think we can do only best-effort approach. Here are some reasons
 * We are able to figure out only the configured limit for receiving. `ConcurrencyLimit` does not prevent users from sending more messages concurrently outside of handler
 * We are able to do only the guestimation to check if the MaxPoolSize has a proper value. For example if  concurrency is 100 and each receive sends message as isolated dispatches s/he will need ~200 open connections. There are also other factors to consider such as custom-checks, business data sql connections and execution environment in general.

I would propose to
 * Log `WARN` whenever user does not specify the MaxPoolSize option (this does not cover custom connection factory though)
 * Consider providing better [documentation](http://docs.particular.net/nservicebus/sqlserver/configuration#connection-strings) regarding how to arrive at proper pool size

HEskandari:
What is the scope of this? Should it be backported to v2 as well?

tmasternak:
@HEskandari this is v3 only issue. The reason is that by default in v5 of core concurrency limit is set to 1 which makes running out of connections impossible. With v6  concurrency limit defaulting to 100 the possibility of that happening **was** quite  big. The default has been [changed](https://github.com/Particular/NServiceBus/issues/4089) to be Max(2, LogicalProcessorCount).

That being said the scope of the issue is to put a warning just as it said in description. Please disregard my later comment https://github.com/Particular/NServiceBus.SqlServer/issues/204#issuecomment-212392885.

HEskandari:
Think we should also update all the connection strings in the snippets on [this page](http://docs.particular.net/nservicebus/sqlserver/connection-settings) to give people reading the doco the right impression. 

Should I add it to the PoA?

SzymonPobiega:
:+1: for adding pool size to all connection strings in context of sql transport.

tmasternak:
[...]

PR #297, opened 2016-09-06
Added connection pool validation. Related to #204.

HEskandari:
This will check connection pool settings on the provided connection string for send and receive infrastructures.

### DIFF
--- a/src/NServiceBus.SqlServer.UnitTests/ConnectionPoolValidatorTests.cs
+++ b/src/NServiceBus.SqlServer.UnitTests/ConnectionPoolValidatorTests.cs
@@ -9,49 +9,49 @@
         [Test]
         public void Is_not_validated_when_connection_pooling_not_specified()
         {
-            var valid = ConnectionPoolValidator.Validate("Initial Catalog = xxx");
+            var result = ConnectionPoolValidator.Validate("Initial Catalog = xxx");
 
-            Assert.IsFalse(valid);
+            Assert.IsFalse(result.IsValid);
         }
 
         [Test]
         public void Is_validated_when_both_min_and_max_pool_size_is_specified()
         {
-            var valid = ConnectionPoolValidator.Validate("Initial Catalog = xxx; min pool size = 20; max pool size=120");
+            var result = ConnectionPoolValidator.Validate("Initial Catalog = xxx; min pool size = 20; max pool size=120");
 
-            Assert.IsTrue(valid);
+            Assert.IsTrue(result.IsValid);
         }
 
         [Test]
         public void Is_not_validated_when_only_min_pool_size_is_specified()
         {
-            var valid = ConnectionPoolValidator.Validate("Initial Catalog = xxx; Min Pool Size = 20;");
+            var result = ConnectionPoolValidator.Validate("Initial Catalog = xxx; Min Pool Size = 20;");
 
-            Assert.IsFalse(valid);
+            Assert.IsFalse(result.IsValid);
         }
 
         [Test]
         public void Is_not_validated_when_only_max_pool_size_is_specified()
         {
-            var valid = ConnectionPoolValidator.Validate("Initial Catalog = xxx; Max Pool Size = 200");
+            var result = ConnectionPoolValidator.Validate("Initial Catalog = xxx; Max Pool Size = 200");
 
-            Assert.IsFalse(valid);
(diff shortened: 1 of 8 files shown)


HEskandari:
@Particular/sqlserver-transport-maintainers Please check my last comments on #204 first and then do the review.

src/NServiceBus.SqlServer/Configuration/ConnectionPoolValidator.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace NServiceBus.Transport.SQLServer
+{
+    using System.Data.Common;
+    using System.Data.SqlClient;
+    using Logging;
+
+    class ConnectionPoolValidator
+    {
+        public static bool Validate(string connectionString)
```
SzymonPobiega: Why does this return `bool`? It seems to be not used by the calling code.

HEskandari: The bool result is currently used for testing.

Initially I was thinking to use something like `StartupCheckResult` but there's only success and fail and we need to `WARN` here so rolled that back. I can change it so that the Validate method returns a <bool, string> and then do the logging outside if return false. That way it'd behave more like `ConnectionStringsValidator`. What do you think @SzymonPobiega?

src/NServiceBus.SqlServer.UnitTests/ConnectionPoolValidatorTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        [Test]
+        public void Is_not_validated_when_only_max_pool_size_is_specified()
+        {
+            var valid = ConnectionPoolValidator.Validate("Initial Catalog = xxx; Max Pool Size = 200");
+
+            Assert.IsFalse(valid);
+        }
+
+        [Test]
+        public void Is_not_validated_when_pooling_is_disabled()
```
tmasternak: I think the purpose of validation is to make sure that users will not run out of connection due to reaching connection pool limit. With pooling disabled that is not possible right?
If so should we treat connection strings with disabled pooling as valid?

src/NServiceBus.SqlServer/Configuration/ConnectionPoolValidator.cs
tmasternak: I am +1 for moving the logging outside of `ConnectionsPoolValidator` - it will make it more single-purpose.

HEskandari:
@Particular/sqlserver-transport-maintainers please do another review. All the issues should have been addressed.

src/NServiceBus.SqlServer/Configuration/SqlConnectionFactory.cs
```diff
@@ -11,15 +12,21 @@ public SqlConnectionFactory(Func<Task<SqlConnection>> factory)
             openNewConnection = factory;
         }
 
-        public Task<SqlConnection> OpenNewConnection()
+        public async Task<SqlConnection> OpenNewConnection()
         {
-            return openNewConnection();
+            var connection = await openNewConnection().ConfigureAwait(false);
+
+            ValidateConnectionPool(connection.ConnectionString);
```
SzymonPobiega: What's the perf impact of parsing the connection string every single time? Should we keep a cache keyed by connection string to avoid validating it every single time?

HEskandari: The function itself is doing the check to run / log only once. Isn't that enough?

SzymonPobiega: Isn't that called each time a new connection is created?

HEskandari: Did a quick check...the amount of time it takes to validate a connection string for one million time is less than 3 seconds.

HEskandari: @SzymonPobiega Also, the flag is a static variable which is shared across factory instances, so it will be checked only once. This was not the case for `LegacySqlConnectionFactory` which I've fixed.

SzymonPobiega: Sorry, I missed that flag. :+1:

PR #307, opened 2016-10-05
[SPIKE] Native timeouts

SzymonPobiega:
Native (not requiring persistence) implementation of delayed delivery
 * Uses a table for each endpoint (with `.Delayed` suffix)
 * Delayed messages destined to other endpoint have their ultimate destination in the `Destination` column
 * Delayed messages which mature (their due date has come) are moved periodically (by default every 5 seconds) to the input queue
 * Received messages with non-null `Destination` are immediately forwarded to their ultimate destination instead of processing them in the pipeline

### Comparison to NHibernate peristence based delayed messages

An immediate send cost is 2 (in terms of DB operations). The added cost of a delayed send when using non-native delayed delivery is 8 DB ops. The added cost of a delayed send to self (most common scenario) is 1 DB op (the move operation) and that operation is executed once no matter how many timeouts have matured. Even when considering the most pessimistic scenario when it moves a single message, we reduce the added cost of delayed delivery roughly 8 times.

#### Immediate send

1. Send to destination queue (`INSERT`)
2. Receive at destination queue (`DELETE`)

#### NHibernate

1. Send to satellite (`INSERT`)
2. Receive in satellite (`DETELE`)
3. Store timeout (`INSERT`)
4. Query timeouts (`SELECT`)
5. Send dispatch request (`INSERT`)
6. Receive dispatch request in satellite (`DETELE`)
7. Query timeout (`SELECT`)
8. Dispatch message (`INSERT`)
9. Remove timeout (`DELETE`)
10. Receive at destination queue (`DELETE`)

#### Native (same endpoint)

1. Send to timeout queue (`INSERT`)
2. Move to input queue when matured (`DELETE` + `INSERT`)
3. Receive at destination queue (`DELETE`)

#### Native (different endpoint)

1. Send to timeout queue (`INSERT`)
2. Move to input queue when matured (`DELETE` + `INSERT`)
3. Receive at local input queue (`DELETE`)
4. Forward to destination queue (`INSERT`)
5. Receive at destination queue (`DELETE`)

### DIFF
--- a/src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
+++ b/src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
@@ -36,7 +36,7 @@ namespace NServiceBus.Transport.SQLServer
             }
             catch (Exception ex)
             {
-                ThrowFailedToSendException(ex);
+                throw new Exception("Failed to store a delayed message.", ex);
             }
         }
 
@@ -48,11 +48,6 @@ namespace NServiceBus.Transport.SQLServer
             {
                 return command.ExecuteNonQueryAsync();
             }
-        }
-
-        void ThrowFailedToSendException(Exception ex)
-        {
-            throw new Exception($"Failed to store a delayed message.", ex);
         }
 
         public async Task<int> Purge(SqlConnection connection)
--- a/src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
+++ b/src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
@@ -1,0 +1,73 @@
+namespace NServiceBus.Transport.SQLServer
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Threading.Tasks;
+    using Transport;
+
+    class DelayedMessageTable
+    {
+        public DelayedMessageTable(string inputQueue, string delayedMessageTableSuffix, QueueAddressParser addressParser)
+        {
+            using (var sanitizer = new SqlCommandBuilder())
+            {
+                var inputQueueAddress = addressParser.Parse(inputQueue);
+
+                inputQueueTable = sanitizer.QuoteIdentifier(inputQueueAddress.TableName);
+                schema = sanitizer.QuoteIdentifier(inputQueueAddress.SchemaName);
+
+                delayedStoreTable = sanitizer.QuoteIdentifier(inputQueueAddress.TableName + "." + delayedMessageTableSuffix);
+            }
+        }
+
+        public async Task Store(OutgoingMessage message, DateTime dueTime, string destination, SqlConnection connection, SqlTransaction transaction)
+        {
+            var messageRow = DelayedMessageRow.From(message.Headers, message.Body, dueTime, destination);
+
+            var commandText = string.Format(Sql.StoreDelayedMessageText, schema, delayedStoreTable);
+
+            try
+            {
+                using (var command = new SqlCommand(commandText, connection, transaction))
+                {
+                    messageRow.PrepareSendCommand(command);
+                    await command.ExecuteNonQueryAsync().ConfigureAwait(false);
+                }
+            }
+            catch (Exception ex)
+            {
+                ThrowFailedToSendException(ex);
+            }
+        }
+
+        public Task MoveMaturedMessages(SqlConnection connection, SqlTransaction transaction)
+        {
+            var commandText = string.Format(Sql.MoveMaturedDelayedMessageText, schema, delayedStoreTable, inputQueueTable);
+
+            using (var command = new SqlCommand(commandText, connection, transaction))
+            {
+                return command.ExecuteNonQueryAsync();
+            }
+        }
+
+        void ThrowFailedToSendException(Exception ex)
+        {
+            throw new Exception($"Failed to store a delayed message.", ex);
+        }
+
+        public async Task<int> Purge(SqlConnection connection)
+        {
+            var commandText = string.Format(Sql.PurgeText, schema, delayedStoreTable);
+
+            using (var command = new SqlCommand(commandText, connection))
+            {
+                return await command.ExecuteNonQueryAsync().ConfigureAwait(false);
+            }
+        }
+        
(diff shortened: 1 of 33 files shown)


SzymonPobiega:
@Particular/sqlserver-transport-maintainers please take a look

tmasternak:
👍 Good work!

I think this proves it will work and we need to figure out/discuss details.

src/NServiceBus.SqlServer/Receiving/ReceiveWithNativeTransaction.cs
```diff
@@ -73,7 +73,15 @@ public ReceiveWithNativeTransaction(TransactionOptions transactionOptions, SqlCo
                 return null;
             }
 
-            return receiveResult.Message;
+            var message = receiveResult.Message;
```
tmasternak: I think this might not be good place to put forwarding logic. It makes `ReceiveStrategy` more complicated and it's already hard to reason about. E.g. it was not obvious for me that this will work correctly with recoverability.

What if we tweaked `onMessage` delegate and put the logic of pushing message to full pipeline or passing to `MessageDispatcher` there?

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
```diff
@@ -21,21 +22,60 @@ public MessageDispatcher(IQueueDispatcher dispatcher, QueueAddressParser address
...
             }
-            return dispatchMethod(deduplicatedOperations);
+            return dispatchMethod(deduplicatedOperations.Values.ToArray());
         }
 
         IQueueDispatcher dispatcher;
         QueueAddressParser addressParser;
-        static OperationByMessageIdAndQueueAddressComparer OperationByMessageIdAndQueueAddressComparer = new OperationByMessageIdAndQueueAddressComparer();
+
+        sealed class DeduplicationKey
```
tmasternak: This is probably not needed for this PR but I like it ;).

src/NServiceBus.SqlServer/Sending/TableBasedQueueDispatcher.cs
```diff
@@ -88,13 +95,43 @@ public TableBasedQueueDispatcher(SqlConnectionFactory connectionFactory)
...
+                else
+                {
+                    var queueAddress = addressParser.Parse(operation.Destination);
+                    var queue = new TableBasedQueue(queueAddress);
+                    await queue.Send(operation.Message, connection, transaction).ConfigureAwait(false);
+                }
+            }
+        }
+
+        static DateTime? GetDueTime(IOutgoingTransportOperation operation)
```
tmasternak: Extension method?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -21,17 +22,25 @@ internal SqlServerTransportInfrastructure(QueueAddressParser addressParser, Sett
...
         /// </summary>
-        public override IEnumerable<Type> DeliveryConstraints { get; } = new[]
+        public override IEnumerable<Type> DeliveryConstraints
         {
-            typeof(DiscardIfNotReceivedBefore)
-        };
+            get
+            {
+                yield return typeof(DiscardIfNotReceivedBefore);
+                if (settings.HasSetting<DelayedDeliverySettings>())
```
tmasternak: Why this is needed?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            var commandText = string.Format(Sql.MoveMaturedDelayedMessageText, schema, delayedStoreTable, inputQueueTable);
+
+            using (var command = new SqlCommand(commandText, connection, transaction))
+            {
+                return command.ExecuteNonQueryAsync();
+            }
+        }
+
+        public async Task<int> Purge(SqlConnection connection)
```
tmasternak: This is not used anywhere and probably should get called when input queue get's purged on startup.

SzymonPobiega: Well, I was thinking about this. If you use native timeouts, you expect them to be deleted when the input queue is purged. On the other hand, if you use the Timeout Manager with persistence, purging queue on startup does not clean the queues...

src/NServiceBus.SqlServer/Receiving/ReceiveWithNativeTransaction.cs
SzymonPobiega: I'll try.

src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs
SzymonPobiega: Well, I had to change the way operations are passed around because now I also need constraints in the lower layers, that's why I introduced this guy.

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
SzymonPobiega: Because the native delayed delivery is opt-in. If you don't configure it explicitly (which adds `DelayedDeliverySettings` to the settings), the transport does not report it handles these constraints. This is in line for ASQ (cc @Scooletz ) in the way we want to release opt-in native timeout support in a minor and then switch completely in major.

Scooletz:
It's quite interesting that implementing the same feature in two different transports we take two different paths. In `ASQ` we use our `Dispatcher` directly for timeouts, sending the message to the initial destination when the time comes. In `SQL` you push the message through the pipeline.

Having this implemented a bit differently it's not a disadvantage though.

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
tmasternak: That makes sense. Thx for clarification.

MarcinHoppe:
Spike was successful. We now know we can implement native delayed delivery in SQL Server transport.

Closing the PR but leaving the branch for future reference.

Issue #308, opened 2016-10-05
Possibility of deadlocks when scaled out


ramonsmits:
Running the SQL transport scaled out can result in deadlocks:
## Deadlock graph

![image](https://cloud.githubusercontent.com/assets/152998/19114770/b4284d6c-8b0f-11e6-95fc-95c3d2477532.png)
## Exception

NServiceBus.StartableEndpoint:

```
2016-10-05 13:12:07.8822|FATAL|16|NServiceBus.StartableEndpoint|Receiver Main failed to initialize.
System.Data.SqlClient.SqlException (0x80131904): Transaction (Process ID 647) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlCommand.InternalEndExecuteNonQuery(IAsyncResult asyncResult, String endMethod, Boolean isInternal)
   at System.Data.SqlClient.SqlCommand.EndExecuteNonQueryInternal(IAsyncResult asyncResult)
   at System.Data.SqlClient.SqlCommand.EndExecuteNonQueryAsync(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NServiceBus.Transport.SQLServer.TableBasedQueue.<Purge>d__12.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #314, opened 2016-10-10
[WIP] Better TTBR query

tmasternak:
## Changes
This is a proposal of a change to the query used to purging expired messages from input queue. The changes were made only to the PurgeBatchOfExpiredMessages query and include:
 * Inner select:
  * Running in `READ UNCOMMITTED` (`NOLOCK` hint) and without `ORDER BY` on `RowVersion` column
  * Returns `RowVersion` instead of `Id`
 * Outer delete
  * `WHERE` clause includes only `RowVersion`

## Why
Direct reason for changes was #308. Which reported a deadlock situation that was caused by `receiveMessage` and `purgeBatchOrExpiredMessages` queries running at the same time. Here is the diagram that was included in the original issue.
image
Process on the left is a `readMessage` operation that tries to delete row from input queue. To do that it needs exclusive IX lock both on clustered index `on RowVersion` column and non-clustered index `Expired` column. As seen on the diagram it already has IX on the second and waits for the first one.
Process on the right is a `purgeBatchOrExpiredMessages` operation that runs in `READ COMMITTED` isolation level and holds shared U lock on non-clustered index (caused by `WHERE [Expires] < GETUTCDATE()` in inner select) and tries to get another shared U lock on clustered index to perform `ORDER by RowVersion`. This causes a deadlock. The root cause of deadlock is the fact that locks are taken in reverse order by `readMessage` and `purgeBatchOrExpiredMessages`.

The changes proposed in this PR do not resolve the problem completely but I believe make the likelihood of deadlock smaller. Here is why: 
 * Running inner select in `purgeBatchOrExpiredMessages` prevents it from holding any locks 
 * Returning `RowVersion` from inner select which is clustered index encourages using clustered index and holding locks only on rows which are to be deleted. Screenshot below shows query plans for new query (first one) and the old query (lower one). "1" shows the inner select part and "2" loop that uses results of inner select to find rows in clustered index by `RowVersion`

image

This change puts us in situation when deadlocking between two queries could happen when they they try to delete the same row. That being said this should be a border case scenario because both queries use disjoint predicates in where statements. When it comes to scaled-out instance the queries for reads and purging will most of the times use the same execution plans which will prevent deadlocks as well. 

/cc: @Particular/sqlserver-transport-maintainers @ramonsmits @Scooletz

### DIFF
--- a/src/NServiceBus.SqlServer/Queuing/Sql.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Sql.cs
@@ -66,7 +66,7 @@ namespace NServiceBus.Transport.SQLServer
                     EXEC sp_releaseapplock @Resource = '{0}_{1}_lock'
                   END";
 
-        internal const string PurgeBatchOfExpiredMessagesText = "DELETE FROM {1}.{2} WHERE [RowVersion] IN (SELECT TOP ({0}) [Id] FROM {1}.{2} WITH (NOLOCK) WHERE [Expires] < GETUTCDATE())";
+        internal const string PurgeBatchOfExpiredMessagesText = "DELETE FROM {1}.{2} WHERE [RowVersion] IN (SELECT TOP ({0}) [RowVersion] FROM {1}.{2} WITH (NOLOCK) WHERE [Expires] < GETUTCDATE())";
 
         internal const string CheckIfExpiresIndexIsPresent = @"SELECT COUNT(*) FROM [sys].[indexes] WHERE [name] = '{0}' AND [object_id] = OBJECT_ID('{1}.{2}')";
 
--- a/src/NServiceBus.SqlServer/Queuing/Sql.cs
+++ b/src/NServiceBus.SqlServer/Queuing/Sql.cs
@@ -66,7 +66,7 @@ namespace NServiceBus.Transport.SQLServer
                     EXEC sp_releaseapplock @Resource = '{0}_{1}_lock'
                   END";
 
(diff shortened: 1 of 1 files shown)


Scooletz:
I like the described approach.
If we had an SQL instance with a lot of traffic (like the one from End2End tests), we could try to run this query in there as well. Or exports statistics and run it locally. The last think we want is to run this query with a different behavior on the production environment.

src/NServiceBus.SqlServer/Queuing/Sql.cs
```diff
@@ -66,7 +66,7 @@ [Expires] ASC
                     EXEC sp_releaseapplock @Resource = '{0}_{1}_lock'
                   END";
 
-        internal const string PurgeBatchOfExpiredMessagesText = "DELETE FROM {1}.{2} WHERE [Id] IN (SELECT TOP ({0}) [Id] FROM {1}.{2} WITH (UPDLOCK, READPAST, ROWLOCK) WHERE [Expires] < GETUTCDATE() ORDER BY [RowVersion])";
+        internal const string PurgeBatchOfExpiredMessagesText = "DELETE FROM {1}.{2} WHERE [RowVersion] IN (SELECT TOP ({0}) [Id] FROM {1}.{2} WITH (NOLOCK) WHERE [Expires] < GETUTCDATE())";
```
ramonsmits: Instead of the '(NO LOCK)' shouldn't the transaction isolation level be set to `READ UNCOMMITTED`?

ramonsmits: @tmasternak Agreed its the same behavior but I personally don't like to prevent the use of hints in queries.

SzymonPobiega:
LGTM, merging.

PR #330, opened 2016-12-28
Multi catalog

SzymonPobiega:
Implements multi-catalog using `table@[schema]@[catalog]` address format that is backwards-compatible with pre-3 and 3.0.0 transports. 

TODO:
 * [x] ~Add wire-compatibility tests to `EndToEnd`~ Adding tests is handled by separate issue #353 
 * [x] Decide if we want to support new catalog part in LegacyMultiInstanceMode - we don't
 * [ ] Discuss validation of multi-instance connection string


SzymonPobiega:
@MarcinHoppe @HEskandari please take a look. I removed the assertion from the SQL injection via endpoint reply address test because I could not make it work.

SzymonPobiega:
Also I am not sure how to test the wire-level compat with 2.X and 3.0 with regards to the address form. Should I copy the 3.0 address parting code to a test and verify it works against addresses containing the catalog name?

MarcinHoppe:
Maybe we could add tests to cover multi-* addressing here:

https://github.com/Particular/EndToEnd/tree/master/src/TransportCompatibilityTests/SqlServerV3

?

SzymonPobiega:
@MarcinHoppe sounds like a plan. But we need to at least have a 3.1 package on myget to get these tests to work, right?

MarcinHoppe:
AFAIK yes.

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,135 @@
...
     using System.Data.SqlClient;
+    using System.Linq;
 
     class QueueAddress
     {
-        public QueueAddress(string tableName, string schemaName)
+        public QueueAddress(string catalog, string schemaName, string tableName)
         {
             Guard.AgainstNullAndEmpty(nameof(tableName), tableName);
+            if (catalog == "")
```
HEskandari: String.Empty? Should we also handle case of a whitespace only string?

src/NServiceBus.SqlServer.UnitTests/QueueAddressTests.cs
```diff
@@ -7,32 +7,39 @@
...
-        [TestCase("my.table@[my.schema]", "my.table", "my.schema")]
-        public void Should_parse_address(string transportAddress, string expectedTableName, string expectedSchema)
+        [TestCase("table@schema", "table", "schema", null)]
+        [TestCase("my.table@my.schema", "my.table", "my.schema", null)]
+        [TestCase("table", "table", null, null)]
+        [TestCase("table@[my.schema]", "table", "my.schema", null)]
+        [TestCase("my.table@[my.schema]", "my.table", "my.schema", null)]
+        [TestCase("my.table@[my.schema]@[my.catalog]", "my.table", "my.schema", "my.catalog")]
+        [TestCase("my.table@my.schema@my.catalog", "my.table", "my.schema", "my.catalog")]
+        [TestCase("my.table@[]@[my.catalog]", "my.table", "", "my.catalog")]
```
HEskandari: I would expected this format to fail (just my opinion though).

SzymonPobiega: The last one is the way to say: send to `my.table` in `my.catalog` using default configured schema. We can't get rid of `[]` (or just space) because it would be interpreted as `my.table` in `my.catalog` schema in the current catalog.

SzymonPobiega:
@Particular/sqlserver-transport-maintainers ping.

SzymonPobiega:
@Particular/sqlserver-transport-maintainers ping :-)

HEskandari:
@SzymonPobiega Is there anything else left here? I thought it was already reviewed?

SzymonPobiega:
@HEskandari I was waiting for the final `LGTM`s from you and @MarcinHoppe :-)

MarcinHoppe:
On it now @SzymonPobiega...

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
```diff
@@ -62,6 +62,25 @@ public static TransportExtensions<SqlServerTransport> UseSchemaForQueue(this Tra
         }
 
         /// <summary>
+        /// Specifies custom schema for given endpoint.
+        /// </summary>
+        /// <param name="transportExtensions">The <see cref="TransportExtensions{T}" /> to extend.</param>
+        /// <param name="endpointName">Endpoint name.</param>
+        /// <param name="catalog">Custom catalog value.</param>
+        /// <returns></returns>
+        public static TransportExtensions<SqlServerTransport> UseCatalogForEndpoint(this TransportExtensions<SqlServerTransport> transportExtensions, string endpointName, string catalog)
```
MarcinHoppe: Do we also need to specify it for queues and not endpoints (e.g. audit, error queue)?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -155,15 +156,38 @@ public override TransportSubscriptionInfrastructure ConfigureSubscriptionInfrast
...
+                if (parser.TryGetValue("Initial Catalog", out catalog) || parser.TryGetValue("Database", out catalog))
+                {
+                    result = result.SetProperty(SettingsKeys.CatalogPropertyKey, (string) catalog);
+                }
+                else
+                {
+                    throw new Exception("Initial Catalog property is mandatory in the connection string.");
+                }
+            }
+            else if (!settings.HasSetting("PublicReturnAddress"))
```
MarcinHoppe: Is this condition related? If not, should go into a separate issue and PR.

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,147 @@
...
 {
+    using System;
+    using System.Collections.Generic;
     using System.Data.SqlClient;
+    using System.Linq;
 
     class QueueAddress
     {
-        public QueueAddress(string tableName, string schemaName)
+        public QueueAddress(string catalog, string schemaName, string tableName)
```
MarcinHoppe: `catalogName` would be more consistent with `schemaName` and `tableName`?

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,147 @@
...
         {
             Guard.AgainstNullAndEmpty(nameof(tableName), tableName);
+            if (catalog != null && IsEmptyOrWhiteSpace(catalog))
```
MarcinHoppe: Can this be shortened to a call to `String.IsNullOrWhiteSpace`?

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,147 @@
...
+            {
+                throw new ArgumentException("Catalog name cannot be empty or whitespace.", nameof(catalog));
+            }
+            if (catalog != null && schemaName == null)
+            {
+                throw new Exception("If catalog is specified, schema name has to be specified too.");
```
MarcinHoppe: Why? Can't we assume a default user schema for a given catalog?

MarcinHoppe: Or even `dbo`?

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,147 @@
...
         }
 
+        public string Catalog { get; }
         public string TableName { get; }
         public string SchemaName { get; }
+        public string Unquoted { get; }
+        public string Quoted { get; }
+        public bool HasDefinedSchema => !string.IsNullOrWhiteSpace(SchemaName);
 
         public static QueueAddress Parse(string address)
```
MarcinHoppe: I know this might sound like adding another problem but... regular expressions?

src/NServiceBus.SqlServer.AcceptanceTests/When_ReplyTo_address_does_not_exist.cs
```diff
@@ -25,7 +25,6 @@ await Scenario.Define<Context>()
             var failedMessage = exception.FailedMessages.Single();
 
             Assert.That(failedMessage.Exception.Message, Contains.Substring("Failed to send message to"));
-            Assert.That(failedMessage.Exception.Message, Contains.Substring("error]] VALUES(NEWID(), NULL, NULL, 1, NULL, '', NULL); DROP TABLE [Victim]]; INSERT INTO [error"));
```
MarcinHoppe: What's the reason for removing this line?

src/NServiceBus.SqlServer.AcceptanceTests/ConfigureEndpointSqlServerTransport.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Data.SqlClient;
 using System.Linq;
+using System.Runtime.InteropServices;
```
MarcinHoppe: The change does not seem to use anything from this namespace?

src/NServiceBus.SqlServer.AcceptanceTests/MultiCatalog/When_using_multicatalog.cs
```diff
@@ -0,0 +1,96 @@
...
+        public class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.UseTransport<SqlServerTransport>()
+                        .ConnectionString(ReceiverConnectionString)
+                        .UseSchemaForEndpoint(ReceiverEndpoint, "receiver")
+                        .UseCatalogForEndpoint(SenderEndpoint, "nservicebus1");
```
MarcinHoppe: Why does the receiver has to know the sender catalog? Shouldn't it come from the return address?

src/NServiceBus.SqlServer.AcceptanceTests/When_ReplyTo_address_does_not_exist.cs
SzymonPobiega: I left you a comment here https://github.com/Particular/NServiceBus.SqlServer/pull/330#issuecomment-269943264. Basically I could not figure out what should be the expression that simulates the injection attack now that we have better address encapsulation (the quotes are no longer hard coded in the SQL template).

src/NServiceBus.SqlServer.AcceptanceTests/MultiCatalog/When_using_multicatalog.cs
SzymonPobiega: Removed.

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
SzymonPobiega: Nope. This is an equivalent of `IsNotNullAnd(EmptyOrWhiteSpace)` :( Meaning, it can be null, but if it is not null, then it cannot be whitespace.

SzymonPobiega:
@MarcinHoppe while trying to apply your suggestions I got a bit stuck. Will come back to this after my vacation.

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,147 @@
...
+            }
+            if (catalog != null && schemaName == null)
+            {
+                throw new Exception("If catalog is specified, schema name has to be specified too.");
+            }
+            Catalog = Unquote(catalog);
+            TableName = tableName;
+            SchemaName = Unquote(schemaName);
+            Quoted = string.Join(".", GetAllParts().Select(QuoteIdentifier));
+            Unquoted = string.Join(".", GetAllParts());
```
tmasternak: Does it really belong here? It's used for table creation lock and logs. The first could be moved to QueueCreator (the logic of removing unsupported characters). Do you see any problems with having quoted identifiers in log statements?

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,147 @@
...
+        }
 
-            TableName = UnescapeIdentifier(tableName);
-            SchemaName = UnescapeIdentifier(schemaName);
+        static bool IsEmptyOrWhiteSpace(string value)
```
tmasternak: Why not Linq and inlining in the if?

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,147 @@
...
+        public string Unquoted { get; }
+        public string Quoted { get; }
+        public bool HasDefinedSchema => !string.IsNullOrWhiteSpace(SchemaName);
 
         public static QueueAddress Parse(string address)
         {
-            if (address.Contains("@"))
+            var location = address;
+
+            var rightBrackets = 0;
```
tmasternak: Do we need to count brackets? How about we scan the string from right-to-left and have a bool flat that tracks if we are in inside `[]`. If yes than we ignore any `@` otherwise we treat is as part separator?

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -1,54 +1,147 @@
 ﻿namespace NServiceBus.Transport.SQLServer
 {
+    using System;
+    using System.Collections.Generic;
     using System.Data.SqlClient;
+    using System.Linq;
 
     class QueueAddress
```
tmasternak: Do we lack a concept here? Should our types reflect two cases? Something like
 1. A result of address parsing that might not be fully specified This is the current `QueueAddress`.
 2. A fully specified address. Maybe `CanonicalAddress` that has all three parts specified.

In such case we could have a well defined areas of code that expect `CanonicalAddress` and  a `Parser` that would translate from `QueueAddress` to `CanonicalAddress` using configuration and defaults.

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -155,15 +156,38 @@ public override TransportSubscriptionInfrastructure ConfigureSubscriptionInfrast
...
             string schema;
             if (schemaSettings.TryGet(instance.Endpoint, out schema) == false)
             {
                 schema = addressParser.DefaultSchema;
             }
 
-            return instance.SetProperty(SettingsKeys.SchemaPropertyKey, schema);
+            var result = instance.SetProperty(SettingsKeys.SchemaPropertyKey, schema);
+
+            if (!settings.HasSetting(SettingsKeys.ConnectionFactoryOverride) && !settings.HasSetting(SettingsKeys.LegacyMultiInstanceConnectionFactory))
```
tmasternak: Do we need the `LegacyMultiInstanceConnectionFactory` check here. `SqlServerTransportInfrastructure` will not be used in Legacy Mode.

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
tmasternak: I think we need to do it. Otherwise there would be no way to specify custom catalog for those queues.

SzymonPobiega:
@Particular/sqlserver-transport-maintainers  I've updated the implementation. All tests should be green now.

SzymonPobiega:
@tmasternak can you review https://github.com/Particular/NServiceBus.SqlServer/pull/353 ? This is I think the last requirement to finish this one.

SzymonPobiega:
@MarcinHoppe @MarcinHoppe I've redesigned this a bit. Please review when you have time.

src/NServiceBus.SqlServer/SqlServerTransport.cs
```diff
@@ -45,14 +36,48 @@ bool LegacyMultiInstanceModeTurnedOn(SettingsHolder settings)
...
+            else
+            {
+                var catalog = GetDefaultCatalog(settings, connectionString);
+                var addressParser = new QueueAddressTranslator((string)catalog, "dbo", defaultSchemaOverride, queueSchemaSettings);
+                return new SqlServerTransportInfrastructure(addressParser, settings, connectionString);
+            }
+        }
 
-            return new SqlServerTransportInfrastructure(addressParser, settings, connectionString);
+        static object GetDefaultCatalog(SettingsHolder settings, string connectionString)
```
MarcinHoppe: Why not cast object to string before returning from this method? This way we bother the caller.

src/NServiceBus.SqlServer/Legacy/MultiInstance/LegacySqlConnectionFactory.cs
```diff
@@ -34,7 +32,8 @@ void ValidateConnectionPool(string transportAddress, string connectionString)
                 Logger.Warn(validationResult.Message);
             }
 
-            SetValidated(transportAddress);
+            //TODO: is this correct?
+            //SetValidated(transportAddress);
```
MarcinHoppe: If this is removed then the call to `HasValidated` should also be removed.

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
```diff
@@ -54,9 +54,47 @@ public static TransportExtensions<SqlServerTransport> UseSchemaForQueue(this Tra
...
+        /// <summary>
+        /// Specifies custom schema for given queue.
+        /// </summary>
+        /// <param name="transportExtensions">The <see cref="TransportExtensions{T}" /> to extend.</param>
+        /// <param name="queueName">Queue name.</param>
+        /// <param name="catalog">Custom catalog value.</param>
+        /// <returns></returns>
+        public static TransportExtensions<SqlServerTransport> UseCatalogForQueue(this TransportExtensions<SqlServerTransport> transportExtensions, string queueName, string catalog)
+        {
+            Guard.AgainstNullAndEmpty(nameof(queueName), queueName);
```
SimonCropp: minor but we should guard against null on `transportExtensions`

src/NServiceBus.SqlServer/Legacy/MultiInstance/LegacySqlConnectionFactory.cs
SzymonPobiega: I've fixed it using concurrent dictionary

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
SzymonPobiega: Added more guards :-)

SzymonPobiega:
Rebased against develop and fixed the issues. Ready for another round of reviews.

SzymonPobiega:
I've added the compat tests for the multi-catalog. @tmasternak @MarcinHoppe please re-review this one

SzymonPobiega:
Merged together with native timeouts.

PR #331, opened 2016-12-28
[WIP] Multi-instace via external shovel process

SzymonPobiega:
Extends the multi-catalog support by providing multi-instance support. Catalogs can be declared as remote (need to use outgoing queue). In such case a message is physically send to the outgoing queue instead of the actual destination.

This depends on an external thing (shovel) to move messages from the outgoing queue to the destination queue.

A fully-functional PoC of that shovel thing using `NServiceBus.Raw` is [here](https://github.com/SzymonPobiega/NServiceBus.SqlServer.Shovel). It uses raw receivers to move message around, one by one. A native solution that moves messages in batches can be developed if necessary.


SzymonPobiega:
After thinking a bit about the ideal configuration experience I think we should:
 * Provide a low level API like in this PR (see the demo) that allows to configure all the catalogs/schemas/outgoing queues by hand
 * Add an opt-in feature that would configure this all based on a table in a database which contains row of following structure:
   * QueueName
   * InstanceId
   * Catalog
   * Schema
   * IsActive
 * Add a feature to the shovel that would read the DB metadata and populate such a table. In case of conflicts (same queue define in two different instances/catalogs/schemas) it would create entries with `IsActvice=false`. This would be used e.g. when migrating queues between databases. A table would be considered a queue based on the structure match (same columns/types).

PR #332, opened 2017-01-02
[SPIKE] Multi schema/catalog auto configuration

### LINKED ISSUE #329: Automatically scan database metadata for catalogs and schemas
Allow users to enable automatic address discovery based on the endpoint name. Assuming the table name equals endpoint name, we can browse the database metadata to find the catalog and schema for a given destination queue name. 

This would not work if there are multiple tables with the same name in different schemas and/or catalogs but this is an odd scenario. Generally we require each endpoint to have unique name so unless a user overrides the local address of two endpoints, say `EndpointA` and `EndpointB` to be `someQueue`, this auto-discovery should work. 

Worst case, if user enables it and we discover a duplicate entry, we throw telling her/him to disable the autodiscovery or check why she/he has multiple queue tables with the same name.

SzymonPobiega:
Fixes #329 

New API:

```
config.UseTransport<SqlServerTransport>().AutoDetectQueueLocations(
   scannedSchemas : new string[] {"s1", "s2"},
   scannedCatalogs : new string[] {"cat1", "cat2"});
```

The DB metadata is queried once every minute. Query results are used to update the instances table which contains information about schema and catalog. If there are duplicates in the results (meaning a given queue is present in more than one schema/catalog) an exception is thrown on startup. If a duplicate is detected during runtime, the routing table is not updated to prevent misrouting.


SzymonPobiega:

SzymonPobiega:
I'll move it to labs and continue working on it in my own time.

PR #336, opened 2017-01-18
Fix infinite retries

MarcinHoppe:
Fixes infinite retry loop issue


MarcinHoppe:
@bording @dbelcham @andreasohlund @szymonpobiega Can you take a look?

Issue #319, opened 2016-11-02
What maintenance is required for SQL Transport


boblangley:
What is the recommended database maintenance to keep the SQL transport healthy? Does the customer need to regularly update statistics or similar?

MarcinHoppe:
@boblangley We currently don't have any guidance around that other than to do what should typically be done for an OLTP workload.

Is there anything in particular that you think we should cover in docs?

boblangley:
Do we have specific recommendations for what should typically be done for an OLTP workload, or a reference we can direct users to that describes it? Considering the usage of the transport, while OLTP, is there any of these tasks that we can rule out or highlight as being important for ongoing database health?

MarcinHoppe:
Myself and @boblangley agreed that we will explain our general guidance ("talk to your DBA") in the transport doco.

SzymonPobiega:
@tmasternak and myself did extensive long-term tests with SQL Server transport 6.0 which resulted in releasing 6.0.1 patch that drops the clustered index (in favour of a non-clustered one). We have not seen any performance degradation even during long-term high-throughput tests, indicating that there is no maintenance required to keep SQL Server transport running.

PR #339, opened 2017-01-25
Don’t validate “Min Pool Size” (0 is fine) and give better warning

DavidBoike:
Since there's no reason 0 can't be the minimum, there's no reason to bother the user if this setting isn't defined in the connection string.

### DIFF
--- a/src/NServiceBus.SqlServer.UnitTests/ConnectionPoolValidatorTests.cs
+++ b/src/NServiceBus.SqlServer.UnitTests/ConnectionPoolValidatorTests.cs
@@ -31,14 +31,6 @@
         }
 
         [Test]
-        public void Is_not_validated_when_only_max_pool_size_is_specified()
-        {
-            var result = ConnectionPoolValidator.Validate("Initial Catalog = xxx; Max Pool Size = 200");
-
-            Assert.IsFalse(result.IsValid);
-        }
-
-        [Test]
         public void Is_not_validated_when_pooling_is_enabled_and_no_min_and_max_is_set()
         {
             var result = ConnectionPoolValidator.Validate("Initial Catalog = xxx; Pooling = true");
--- a/src/NServiceBus.SqlServer/Configuration/ConnectionPoolValidator.cs
+++ b/src/NServiceBus.SqlServer/Configuration/ConnectionPoolValidator.cs
@@ -15,7 +15,7 @@
                 return ValidationCheckResult.Valid();
             }
 
-            if (!keys.ContainsKey("Max Pool Size") || !keys.ContainsKey("Min Pool Size"))
+            if (!keys.ContainsKey("Max Pool Size"))
             {
                 return ValidationCheckResult.Invalid(ConnectionPoolSizeNotSet);
             }
@@ -24,7 +24,7 @@
         }
 
         const string ConnectionPoolSizeNotSet = 
-            "Minimum and Maximum connection pooling values are not " +
-            "configured on the provided connection string.";
+            "Maximum connection pooling value (Max Pool Size=N) is not " +
+            "configured on the provided connection string. The default value (100) will be used.";
     }
 }


SimonCropp:
Looks good. Can u also point to the doco that will change based on this behaviour change?

SzymonPobiega:
@SimonCropp I've checked and this PR does not invalidate any doco. That said, I noticed we're using connection strings without the max size value in the doco and that's something to be changed.

SimonCropp:
Inst the fact that we r changing behaviour and no doco needs to change the sign of missing doco?

SzymonPobiega:
Yeah, it is. I'll cover it along with fixing the connection string. What I meant is that @DavidBoike did not make it worse.

SzymonPobiega:
@SimonCropp here's the basic fix https://github.com/Particular/docs.particular.net/pull/2398 but I believe we need a proper one that documents how users should come up with the max pool size value.

MarcinHoppe:
Our general approach so far was not to dive too deep into recommendations as to how configure and manage SQL Server used by the transport, similar to #319.

SzymonPobiega:
@MarcinHoppe minimum what we should probably do is provide users with a formula that allows them to calculate how many connection will NSB need given the settings. We can probably even log this, both in transport and in NH persistence at start.

MarcinHoppe:
@SzymonPobiega I think a better approach would be to teach users how to monitor the number of connection the entire endpoint is using (transport + persistence + user code) and when to adjust it in the connection string (i.e. what symptoms indicate that max number of connections in the pool should be increased). Making developers better.

SzymonPobiega:
@MarcinHoppe shouldn't it be the other way round -- starting from the SQL Server instnace:
 * The DB instance has a hard limit on number of concurrent connections
 * The users should figure out how many processes connect to that instance. Each process should be allocated a portion of that hard connection limit in form of limiting the pool size for it (assuming the process talks to a single instance of SQL server)
 * Given we have a pool size of X, the users should use some kind of formula to figure out what maximum concurrency value he should set for a given endpoint.

MarcinHoppe:
I don't have the DBA knowledge to agree or disagree. I think databases should be tuned by qualified professionals.

SzymonPobiega:
@MarcinHoppe I agree to that. But the DBs job is done when she/he says "Thou shalt not use more than X connections". So the devs put that limit into the connection string and their endpoint blows up because the concurrency limit is too high. But we offer no way for them (other than just guess) to figure out what should be the ratio of `max connections` to `max concurrency`.

MarcinHoppe:
I get it but the answer depends on the traffic patterns and interactions with the persister (timeouts, outbox) and user code. I can't see how we could come up with an analytic formula here.

DavidBoike:
Really? What are the components? I'll make a bunch of guesses here and you all can tell me how I'm ridiculously far off. :wink:

* Message pipeline - each message should have one connection. Everything happening in service to that message should be sharing the same local transaction. Therefore, allow whatever the MaxConcurrency setting is.
* Specific persisters:
  * Subscription lookups - should share the session, or be on a separate database, so 0
  * Saga - shares session - 0
  * Timeouts
    * Setting - shares the session
    * Receiver - Ideally, 1 query should happen at once? I suppose they could overlap, but that doesn't sound like a good situation
  * Outbox - Am I right in saying that using Outbox with SQL transport makes no sense?
  * Gateway - OK I really don't know enough about how the Gateway works here
* User does something completely random in a new transaction on each message on the SAME - OK that's a wildcard, but also sounds pretty unlikely

So by my math, MaxConcurrency plus a small handful (or a 10-20% buffer?) should work, should it not? 

Where are my assumptions wrong?

This sounds like it may have been a problem when the default concurrency in V6 was 100, but now that it's based on logical processors, this seems solvable.

MarcinHoppe:
@SzymonPobiega @DavidBoike I ran the modified [SQL Server Transport and NHibernate Persistence](https://docs.particular.net/samples/sqltransport-nhpersistence/) sample and tried to fiddle with the max pool size. The modification is that the sender publishes around 10 events per second (a publish is followed by a 100ms delay, to be precise). The workload is fairly realistic: it has a saga with timeouts, audits are on and messages end up in the error queue from time to time.

I limited sender concurrency to 2 and max number of connections in the pool to 10. Never saw a problem with this endpoint.

The receiver has the default concurrency (8 on my machine) and I also limited it to 10 connections in the pool. I saw pool timeout exceptions fairly quickly. Then I increased it to 20 and also hit pool timeout exceptions.

**The bad thing is that after some time even a pool of max size 1024 got exhausted.**

I will create an issue for it.

tmasternak:
@DavidBoike here are comments for my end:

Here is what we know can influence the number of connections
  * Message pipeline - not always single connection. When sending Isolated messaged we cannot reuse any existing connections if there is transaction scope or native transaction
  * Outbox - it makes sense to use it with SqlServer if you have separate central transport sql instance and o not want to use DTC. In such case sql connection per message is needed

In addition we have:
  * User opening connection on their own e.g. with Entity Framework
  * CustomChecks or anything else which is not connected to the particular message being processed but runs independently of that
  * If hosted in web application the pool will be shared as well

That being said I do agree that the biggest problem was the old concurrency limit in `Core` and removing `Min Pools Size` constraint is a good move.

DavidBoike:
Yeah, that does make it really hard to make a formula.

SzymonPobiega:
Well, we can at least report on the actual usage, both in SQL transport and in NH persistence. That would give users some clue.

MarcinHoppe:
There are ADO.NET and SQL Server perf counters. Are they not sufficient?

SzymonPobiega:
@MarcinHoppe you're right 👍

PR #351, opened 2017-03-01
Cleanup

SimonCropp:


MarcinHoppe:
@SimonCropp These moves to `Task.WhenAll` look like a good idea but would require more careful connection management (connection per operation) and more connections overall?

SimonCropp:
@MarcinHoppe i reverted the when all. it is outside the scope of "me browsing the code and finding minor things to fix"

PR #353, opened 2017-03-07
Moving transport compat tests to SQL repo

SzymonPobiega:
We need to move the transport compat tests to our repo because:
 * Currently the `EndToEnd` repo needs to be updated each time we release to run tests against most recent versions
 * Because of that there is no easy way to run the tests against PRs or develop
 * `EndToEnd` tests are not run on TC for each build
 * There is much bigger friction in updating the tests in a separate repo

Things done:
 * Moved tests from EndToEnd
 * Refactored the test to be explicit about configurations in each scenario
 * Added more multi-schema tests
 * Added multi-instance tests

TODO
 * [x] Fix failing test
 * [x] Fix style checks on TC 
 * [x] Remove version strings from test
 * [x] Remove test base class - switch to partials also for Callbacks
 * [x] Reformat code


SimonCropp:
so how will this work when we patch an older version? since a patch would need to have changes on two branches

SzymonPobiega:
@SimonCropp we update develop with the latest patch of an old release

SimonCropp:
Don't we have a race condition there? U can't verify and release a patch until it has been verified against develop. U can't update developer until the patch has been released.

SzymonPobiega:
@SimonCropp We can treat this similarly to smoke testing via doco samples. Release on myget, upgrade the tests and verify. For all future releases (3.1 and higher) the tests would already be in the support branches so there would not be the need to go through develop.

SimonCropp:
I get it is possible. But I don't see the value over keeping it in a diff repo. Or diff (dedicated) branch

SzymonPobiega:
@SimonCropp the value is in keeping the code close and making it run as part of the build. So far it lived in a separate repo and has not been run/updated (it even referenced pre-release V6 packages).

SimonCropp:
@SzymonPobiega having to manipulate multiple live branches as part of a releasing an old patch seems wrong. i would prefer the minor friction of having one canonical branch/repo for storing these tests.

SzymonPobiega:
@SimonCropp I think the trade-off is between releasing new versions and releasing patches to existing versions.

If we move it back (or move it to other repo), we would have friction when trying to run these tests against changes we're just making) -- need to create a package, push it to MyGet blablabla

If we keep it close in the repo, we'll have a bit of friction when releasing patches to old versions.

IMO the likelihood of breaking these tests is much bigger when creating new features (because they might be not compatible with old versions) than when we release small bugfixes for critical bugs, that's why I'd like to keep this in the repo.

SimonCropp:
from my perspective i see this causing more long term fiction and pain than the current solution.

so i am a -1

SzymonPobiega:
@SimonCropp the current solution never worked because nobody cared to go there and upgrade the nugets. Let's give this one a try and see if the friction is killing us.

SzymonPobiega:
@tmasternak @MarcinHoppe can you express your opinions here?

MarcinHoppe:
Sorry, I probably don't have enough context, but... why are we even doing this?

SzymonPobiega:
@MarcinHoppe I've updated the description with the reasons for this.

MarcinHoppe:
Do we have a resonably easy way to update thoses tests when the upstream (EndToEnd repo) changes?

SzymonPobiega:
@MarcinHoppe nope. But the actual test cases are actually SQL-specific. The only thing that is/was shared was the infrastructure to run separate app-domains. I don't think there is much value in being able to fetch the changes from the upstream repo is
 * it does not change (recently)
 * the shared code is relatively trivial

MarcinHoppe:
@SzymonPobiega Thanks. Then I think it is good to move them here. Running compat tests on PRs is a big sell for me.

SimonCropp:
doesnt this mean when we have a new support release we need to also update all other support branches?

tmasternak:
@SimonCropp the approach that we want to take is:
 * Compat tests will be added for now to the develop only
 * They will reference the newest version on develop via project reference and rest  of the versions (including newest releases) as nuget packages
 * They will run for the current develop as part of TC build
 * For any releases from will have additional manual step of updating backwards compat test with reference to the new release

The advantage is that we will finally have short feedback loop on develop. The downside is that each release becomes longer but the argument was that, if it becomes painful we will improve it.

After writing this up I do have a question about the evolution of this approach. What happens when we merge the develop to master as part of a release? Do we want to discard the tests? Or do we want to merge them? When would we run them from master?

SzymonPobiega:
@tmasternak we keep them on master in order to make validating hotfixes easier. On TC we always run on develop because it by definition of referencing the latest released version via nuget includes all the stuff there is on master.

The benefit of having them in future versions of master is that, should we have to release a hotfix, we would be able to run the tests for the hotfix directly on a hotfix branch.

tmasternak:
>On TC we always run on develop because it by definition of referencing the latest released version via nuget includes all the stuff there is on master.

So we want to reference the newest version via nuget? I thought it will be project reference.

> The benefit of having them in future versions of master is that, should we have to release a hotfix, we would be able to run the tests for the hotfix directly on a hotfix branch.

I'm not sure about that would be that convenient. Let's say I want to release patch for 3.2 (which has test) and the newest version is 4.2. I would have to update 3.2 branch tests with all versions between 3.2 and 4.2 to check the compatibility with then as well.

SzymonPobiega:
> So we want to reference the newest version via nuget? I thought it will be project reference.

No. I meant from develop we will be able to run all the versions we need.

I don't think we need to update all. Since we are now moving into "latest minor for each supported major" support policy in your case we would only need to update the reference to 4.x. My vision for the flow is:

 * make a fix some branch coming out of `support-X.Y`
 * update the nuget references to other majors (might be 2-3, I don't think more)
 * run the compat tests and all the other tests
 * cherry pick commit to `hotfix-X.Y.Z` branch
 * merge to `support-X.Y` when green

tmasternak:
@SzymonPobiega this is ready to review and merge.

SzymonPobiega:
LGTM.

PR #354, opened 2017-03-08
Clean up sql statements

SimonCropp:


SimonCropp:
I think this one is ready for a review

SzymonPobiega:
@SimonCropp I took liberty to piggy back on this and changed the create table script to reduce nesting via usage of `return` keyword.

PR #356, opened 2017-03-13
Optimize Queueloop

SimonCropp:


MarcinHoppe:
@SimonCropp Is the impact measurable? It seems the I/O will be dominant factor here.

SzymonPobiega:
I like the pre-formatting thing but without further changes it won't give us much because `TableBasedQueue` is created each time we send a message. It should be cached somewhere, reducing further the GC pressure.

SzymonPobiega:
@SimonCropp do you want me to handle the caching of `TableBasedQueue` instances?

SimonCropp:
@SzymonPobiega that would be greatly appreciated

SzymonPobiega:
@tmasternak @MarcinHoppe can you now re-review it?

MarcinHoppe:
@SzymonPobiega Make member variables on `TableBasedQueue` readonly to ensure they are safe to read from concurrently?

SzymonPobiega:
@MarcinHoppe I think our coding standards say we don't use `readonly`

MarcinHoppe:
@SzymonPobiega Where is that stated? Any rationale?

SimonCropp:
Safe to read from concurrently? How does readonly enable that?

MarcinHoppe:
@SimonCropp Strings are immutable and storing references to them in readonly fields would provide protection that these fields are not modified after the object has been created. I think this is a reasonable thing to do to help ensure `TableBasedQueue` behaves correctly when invoked concurrently from different threads.

Is there anything I am missing?

SimonCropp:
in cases like this, where we dont change them, readonly is a self check to remind us that they are scoped per instance and should not be changed. ie the very nature of the fact that we dont change them outside the constructor means they are readonly even without the keyword. if ever the class get so complex that we need readonly to reason about fields scoped per class then that is a sign the class is too complex. essentially all the arguments for private readonly boil down to enabling complexity.

MarcinHoppe:
My intent was to provide a self check, nothing more. If we can have a compiler warn us against introducing a bug, why not use it?

Having said that, I will not fight tooth and nail for this readonly. It's up to @SzymonPobiega to act on my comment or not.

SimonCropp:
@MarcinHoppe so how many time in your career has readonly stopped you from incorrectly setting a field

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
```diff
@@ -51,5 +51,26 @@ static string UnescapeIdentifier(string identifier)
                 return sanitizer.UnquoteIdentifier(identifier);
             }
         }
+
+        bool Equals(QueueAddress other)
+        {
+            return string.Equals(TableName, other.TableName) && string.Equals(SchemaName, other.SchemaName);
```
tmasternak: Are sql identifiers case sensitive?

src/NServiceBus.SqlServer/Queuing/TableBasedQueueCollection.cs
```diff
@@ -0,0 +1,14 @@
+namespace NServiceBus.Transport.SQLServer
+{
+    using System.Collections.Concurrent;
+
+    class TableBasedQueueCollection
```
tmasternak: Is the `Collection` a good name here? Maybe `Builder`?

src/NServiceBus.SqlServer/Queuing/TableBasedQueueCollection.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public TableBasedQueue GetQueue(QueueAddress address)
```
tmasternak: Just `Get` instead of `GetQueue`?

src/NServiceBus.SqlServer/Addressing/QueueAddress.cs
SzymonPobiega: That's a lovely question. The answer is it depends on the collation. In the current code, when deduplicating, we use case-sensitive comparison (https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Sending/OperationByMessageIdAndQueueAddressComparer.cs) which makes sense there.

Here the behavior would be like this: if you pass `MyTable` and `mytable` in two addresses you wound end up with two entries in the cache. If the DB collation is case-sensitive it is what you want. If it is not then you will end up with duplicated cache entries but still everything will be correct.

src/NServiceBus.SqlServer/Queuing/TableBasedQueueCollection.cs
SzymonPobiega: It not only builds but also manages the cache. `Factory`?

SzymonPobiega:
@tmasternak I did the renames.

SzymonPobiega:
This is now green so feel free to hit the green button @tmasternak if you are satisfied with my changes for the stuff you requested.

PR #362, opened 2017-05-11
Use GetOrCreate for endpoint instances

SzymonPobiega:
This removes the blocker for using SQL Server transport in raw mode (dispatcher + pump) as the previous code (Get) assumed the routing feature is enabled.

This does not change the behavior of normal endpoints.


SzymonPobiega:
Removed from the milestone to not include in the release notes.

PR #364, opened 2017-06-05
Native delayed delivery

SzymonPobiega:


src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliverySettings.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace NServiceBus.Transport.SQLServer
+{
+    using System;
+
+    /// <summary>
+    /// Configures native delayed delivery.
+    /// </summary>
+    public class DelayedDeliverySettings
+    {
+        internal string Suffix { get; private set; }
```
SimonCropp: ig they are internal they may as well be props. encapsulation only has value between assembly boundaries

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliverySettings.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        public void DisableTimeoutManager()
+        {
+            TimeoutManagerDisabled = true;
+        }
+
+        /// <summary>
+        /// Configures how often delayed messages are processed (every 5 seconds by default), in seconds.
+        /// </summary>
+        /// <param name="interval"></param>
```
SimonCropp: either document or remove

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliverySettings.cs
```diff
@@ -0,0 +1,45 @@
...
+        public void ProcessingInterval(int interval)
```
SimonCropp: shouldnt this be a timespan?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageRow.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+        
+        static T TryGetHeaderValue<T>(Dictionary<string, string> headers, string name, Func<string, T> conversion)
+        {
+            string text;
+            if (!headers.TryGetValue(name, out text))
+            {
+                return default(T);
+            }
+            var value = conversion(text);
```
SimonCropp: direct return

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                using (var command = new SqlCommand(storeCommand, connection, transaction))
+                {
+                    messageRow.PrepareSendCommand(command);
+                    await command.ExecuteNonQueryAsync().ConfigureAwait(false);
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new Exception("Failed to store a delayed message.", ex);
```
SimonCropp: given `DelayedMessageTable.Store` will already be in the stack trace is there any value add in wrapping the exception?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+
+        public Task MoveMaturedMessages(SqlConnection connection, SqlTransaction transaction)
+        {
+            using (var command = new SqlCommand(moveMaturedCommand, connection, transaction))
+            {
+                return command.ExecuteNonQueryAsync();
```
SimonCropp: generally not awaiting inside a using is a bug.

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
```diff
@@ -38,6 +38,18 @@ public static class SqlConstants
...
 IF (@NOCOUNT = 'OFF') SET NOCOUNT OFF;";
 
+        internal const string StoreDelayedMessageText =
+@"
+    DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
+    IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON'
+    SET NOCOUNT ON;
+
+    INSERT INTO {0} ([Id],[CorrelationId],[ReplyToAddress],[Recoverable],[Expires],[Headers],[Body],[Due])
+    VALUES (@Id,@CorrelationId,@ReplyToAddress,@Recoverable,CASE WHEN @TimeToBeReceivedMs IS NOT NULL THEN DATEADD(ms, @TimeToBeReceivedMs, GETUTCDATE()) END,@Headers,@Body,@Due);;
```
SimonCropp: minor. but i would like some whitespace and newlines to make this more readable for the next person

src/NServiceBus.SqlServer/Sending/TableBasedQueueDispatcher.cs
```diff
@@ -92,12 +98,64 @@ public TableBasedQueueDispatcher(SqlConnectionFactory connectionFactory, QueueAd
...
+
+                string expire;
+                var isTimeoutManagerDelayedMessage = operation.Message.Headers.TryGetValue(TimeoutManagerHeaders.Expire, out expire);
+
+                if (delayedMessageTable == null || (!(due = GetDueTime(operation)).HasValue && !isTimeoutManagerDelayedMessage))
+                {
+                    var address = addressTranslator.Parse(operation.Destination);
+                    var queue = queueFactory.Get(address.QualifiedTableName, address.Address);
+                    await queue.Send(operation.Message.Headers, operation.Message.Body, timeToBeReceived?.MaxTime ?? TimeSpan.MaxValue, connection, transaction).ConfigureAwait(false);
+                }
```
SimonCropp: i would have a return here to reduce nesting

src/NServiceBus.SqlServer/Sending/TableBasedQueueDispatcher.cs
```diff
@@ -92,12 +98,64 @@ public TableBasedQueueDispatcher(SqlConnectionFactory connectionFactory, QueueAd
...
+
+                    if (!isTimeoutManagerDelayedMessage)
+                    {
+                        await delayedMessageTable.Store(operation.Message, due.Value, operation.Destination, connection, transaction).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        //handle old messages delayed with TimeoutManager
+                        var expirationTime = DateTimeExtensions.ToUtcDateTime(expire);
+                        var destination = operation.Message.Headers[TimeoutManagerHeaders.RouteExpiredTimeoutTo];
```
SimonCropp: making `operation.Message.Headers` a local var would make this more readable

src/NServiceBus.SqlServer/Sending/TableBasedQueueDispatcher.cs
```diff
@@ -92,12 +98,64 @@ public TableBasedQueueDispatcher(SqlConnectionFactory connectionFactory, QueueAd
...
+                    // align with TimeoutManager behavior
+                    if (timeToBeReceived != null && timeToBeReceived.MaxTime < TimeSpan.MaxValue)
+                    {
+                        throw new Exception("Delayed delivery of messages with TimeToBeReceived set is not supported. Remove the TimeToBeReceived attribute to delay messages of this type.");
+                    }
+
+                    if (!isTimeoutManagerDelayedMessage)
+                    {
+                        await delayedMessageTable.Store(operation.Message, due.Value, operation.Destination, connection, transaction).ConfigureAwait(false);
+                    }
```
SimonCropp: add a return to reduce nesting

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -18,16 +19,25 @@ internal SqlServerTransportInfrastructure(QueueAddressTranslator addressTranslat
             this.settings = settings;
             this.connectionString = connectionString;
 
-            this.schemaAndCatalogSettings = settings.GetOrCreate<EndpointSchemaAndCatalogSettings>();
+            schemaAndCatalogSettings = settings.GetOrCreate<EndpointSchemaAndCatalogSettings>();
+            delayedDeliverySettings = settings.GetOrCreate<DelayedDeliverySettings>();
 
-            //HINT: this flag indicates that user need to explicitly turn outbox in configuration.
+            //HINT: this flag indicates that user need to explicitly turn outbox in configusration.
```
SimonCropp: configusration?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -114,17 +133,59 @@ ExpiredMessagesPurger CreateExpiredMessagesPurger(SqlConnectionFactory connectio
...
+
+        CanonicalQueueAddress GetDelayedQueueTableName()
+        {
+            if (string.IsNullOrEmpty(delayedDeliverySettings.Suffix))
+            {
+                throw new Exception("Native delayed delivery feature requires configuring a table suffix.");
+            }
+            var delayedQueueLogialAddress = settings.LogicalAddress().CreateQualifiedAddress(delayedDeliverySettings.Suffix);
+            var delayedQueueAddress = addressTranslator.Generate(delayedQueueLogialAddress);
+            var delayedQueueTableName = addressTranslator.GetCanonicalForm(delayedQueueAddress);
```
SimonCropp: direct return

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
SzymonPobiega: I didn't realize that 😮

src/NServiceBus.SqlServer/Sending/TableBasedQueueDispatcher.cs
SzymonPobiega: I'll rewrite this a bit.

SzymonPobiega:
@SimonCropp thanks. I pushed the changes that address your comments.

SzymonPobiega:
🎉

src/NServiceBus.SqlServer.AcceptanceTests/ConfigureEndpointSqlServerTransport.cs
```diff
@@ -25,9 +27,12 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
         connectionString = settings.Get<string>("Transport.ConnectionString");
 
         var transportConfig = configuration.UseTransport<SqlServerTransport>();
-
+		//transportConfig.EnableNativeDelayedMessageDelivery();
```
weralabaj: that doesn't seem to be necessary?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryMessagePump.cs
```diff
@@ -0,0 +1,43 @@
...
+                }
+                await onMessage(context).ConfigureAwait(false);
+            }, context =>
+            {
+                delayedMessageProcessor.HandleError(context);
+                return onError(context);
+            }, criticalError, settings);
+        }
+
+        public void Start(PushRuntimeSettings limitations)
```
weralabaj: limitations ? :)

Why not just simple "settings"?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageProcessor.cs
```diff
@@ -0,0 +1,41 @@
...
+            string forwardDestination;
+            context.Headers.TryGetValue(ForwardHeader, out forwardDestination);
+            if (forwardDestination == null || forwardDestination == localAddress)
+            {
+                context.Headers.Remove(ForwardHeader);
+                //Do not forward the message. Process in local endpoint instance.
+                return false;
+            }
+            var outgoingMessage = new OutgoingMessage(context.MessageId, context.Headers, context.Body);
+            var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(forwardDestination));
```
weralabaj: > new UnicastAddressTag(forwardDestination)

What happens when forwardDestination is null? Is that even a valid scenario? If not why do we need to check for it earlier?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public async Task MoveMaturedMessages(SqlConnection connection, SqlTransaction transaction)
+        {
+            using (var command = new SqlCommand(moveMaturedCommand, connection, transaction))
+            {
+                await command.ExecuteNonQueryAsync().ConfigureAwait(false);
+            }
+        }
+
+        public async Task<int> Purge(SqlConnection connection)
```
weralabaj: that doesn't seem to be used?

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
```diff
@@ -56,6 +85,30 @@ public static class SqlConstants
...
+IF (@NOCOUNT = 'ON') SET NOCOUNT ON;
+IF (@NOCOUNT = 'OFF') SET NOCOUNT OFF;";
+
+        internal const string MoveMaturedDelayedMessageText = @"
+DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
+IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON';
+SET NOCOUNT ON;
+
+WITH message AS (
+    SELECT TOP(100) *
```
weralabaj: I wonder if the number of messages to pick should also be configurable? Then together with processing interval they can be used to fine-tune the endpoint. We could still use 100 as a default. Or would that qualify as a premature optimization and we can introduce later if people request it?

src/NServiceBus.SqlServer/Receiving/QueueCreator.cs
```diff
@@ -29,10 +30,15 @@ public QueueCreator(SqlConnectionFactory connectionFactory, QueueAddressTranslat
                     await CreateQueue(addressTranslator.Parse(sendingAddress), connection, transaction).ConfigureAwait(false);
                 }
 
+                if (delayedMessageTable != null)
```
weralabaj: Can it be null? Why would it be? Does it have to be treated separately from other Receiving side queues or could we try to add it as a standard queue binding, assuming we want to have it always created going forward?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageProcessor.cs
```diff
@@ -0,0 +1,41 @@
+namespace NServiceBus.Transport.SQLServer
+{
+    using System.Threading.Tasks;
+    using Routing;
+
+    class DelayedMessageProcessor
```
weralabaj: General question, may be stupid. Since delayed delivery was/is a Core feature, why native timeouts are not implemented as features as well? Is there some design limitation or  it wouldn't be appropriate for X reasons?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageRow.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static DelayedMessageRow From(Dictionary<string, string> headers, byte[] body, DateTime due, string destination)
+        {
+            var row = new DelayedMessageRow();
+
+            row.id = Guid.NewGuid();
+            headers["NServiceBus.SqlServer.ForwardDestination"] = destination;
+            row.correlationId = TryGetHeaderValue(headers, Headers.CorrelationId, s => s);
+            row.replyToAddress = TryGetHeaderValue(headers, Headers.ReplyToAddress, s => s);
+            row.recoverable = true;
+            row.timeToBeReceived = TryGetHeaderValue(headers, Headers.TimeToBeReceived, s =>
```
weralabaj: since we're validating TTBR can't be used in combination with delayed delivery before calling this method, why do we need to store this header? can we remove that column?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -18,16 +19,25 @@ internal SqlServerTransportInfrastructure(QueueAddressTranslator addressTranslat
...
+        public override IEnumerable<Type> DeliveryConstraints
         {
-            typeof(DiscardIfNotReceivedBefore)
-        };
+            get
+            {
+                yield return typeof(DiscardIfNotReceivedBefore);
+                if (delayedDeliverySettings.TimeoutManagerDisabled)
+                {
+                    yield return typeof(DoNotDeliverBefore);
```
weralabaj: I'm wondering now why is that needed, seems that every time we request delayed delivery for a message then this constraint is added https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryOptionExtensions.cs#L23. Wouldn't that work irrelevant of using or not using timeout manager?

src/NServiceBus.SqlServer.AcceptanceTests/ConfigureEndpointSqlServerTransport.cs
```diff
@@ -25,9 +27,12 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
...
         transportConfig.ConnectionString(connectionString);
 
+        var delayedDeliverySettings = transportConfig.DelayedDelivery();
+        delayedDeliverySettings.TableSuffix("Delayed");
```
danielmarbach: Is table suffix mandatory? If yes why not make it a param of `DelayedDelivery` like `DelayedDelivery(tableSuffix: "Delayed")`

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
SzymonPobiega: If we return that we support the constraint then the TM would not be started. This code is to trick the core that we don't support delayed delivery.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageRow.cs
SzymonPobiega: Good catch. We can remove that column altogether.

src/NServiceBus.SqlServer/Receiving/QueueCreator.cs
SzymonPobiega: The queue table has a different schema so it has to be treated differently. And no, it can't be null in the current code. This is a leftover from the times when native DD was opt-in.

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
SzymonPobiega: Yeah, I agree this should be configurable.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
SzymonPobiega: Probably should be when we purge the queue. Or not? I don't know.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageProcessor.cs
SzymonPobiega: We set it when dispatching. It should never be null

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageProcessor.cs
SzymonPobiega: The general guidance for the transport implementations was to avoid fine-grained features. Transport should not use the feature API, only the transport API. That said, some things cannot be done this way.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryMessagePump.cs
SzymonPobiega: No idea, that was the name copied from the normal `MessagePump`

src/NServiceBus.SqlServer.AcceptanceTests/ConfigureEndpointSqlServerTransport.cs
SzymonPobiega: 👍

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageProcessor.cs
SzymonPobiega: I've added a null check in the send path and another in the receive path that moves messages that (for whatever reason) are lacking that header to the error queue.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageProcessor.cs
SzymonPobiega: I've added a null check in the send path and another in the receive path that moves messages that (for whatever reason) are lacking that header to the error queue.

SzymonPobiega: I actually misunderstood my own code. The header is lacking if the message is not delayed. I've made this more explicit now.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
weralabaj: you mean https://docs.particular.net/nservicebus/messaging/discard-old-messages#discarding-old-messages-at-startup?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
weralabaj: OK, should we be more explicit about the fact this should be removed in future versions?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
SzymonPobiega: Yeah. Should this also purge the delayed messages if they are handled natively? With TM it would not purge them...

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
SzymonPobiega: Well, I expect we remove `TimeoutManagerDisabled` in next version so this will be removed "naturally".

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
weralabaj: I'd say it only makes sense in the dev environment. Contrary to discarding "old" messages in IoT-like  apps, when they are no longer relevant, timeouts rather not should be lost.

weralabaj: So I'm in favour of not purging, similar to what we do for TM.

bording:
Spent some time looking at this, so here are a few comments:

- For the Delay + TTBR exception message, there could be value in aligning with the message that core uses: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/RouteDeferredMessageToTimeoutManagerBehavior.cs#L32. I still need to add the message type to the RabbitMQ error message, because that was a recent change in core.

- There are several TM-related settings that can have some weird interactions when used in conjunction with this native delivery stuff because they all impact if the TM is active or not:

  - Send-only endpoints
  - external timeout manager address
  - Explicitly disabling the feature via DisableFeature

The external address being configured leads to some REALLY weird behavior, so it should never be allowed. The other two end up with the TM disabled but the transport hasn't told core it supports native delivery (via DeliveryConstraints) so things end up in a rather broken state.

Take a look at https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs#L92 to see what I'm talking about. That's done as a startup check.

The other part of that is here: https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs#L43

If either of those are true, then it's safe to go ahead and go into full-native mode. I had originally detected those and threw as well, but it seemed odd to know it was OK to turn off the TM but force the user to add a config setting.

Based on that, there's a case to be made that the DisableTimeoutManager API isn't needed at all, and you could instead just rely on the DisableFeature API. I added DisableTimeoutManager to RabbitMQ before realizing this. It's probably better to add it this point to have a consistent API across transports though...

SzymonPobiega:
@bording thanks for the review! If I read this correctly, the CheckForInvalidSettings will throw if the user used "DisabledFature<TimeoutManager>" rather than going through the proper API. Am I right?

bording:
@SzymonPobiega You're correct. In the general case, it won't be possible to hit that error. When we added tests around the startup check, it seemed wrong that it wouldn't throw on a known bad state, so we left the check in.

While figuring out all the stuff around these settings, I had to examine in detail how features work: when they are created, when they are enabled by default, when the default settings are applied, when prereqs are evaluated, and when setup is run. With the access to settings given at all those various stages, a feature could do some really weird stuff, so that was one more reason to keep that check around.

SzymonPobiega:
@Particular/sqlserver-transport-maintainers please review, this is no longer a WIP. You might first review the multi-catalog PR or review both code changes here.

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -18,16 +20,33 @@ internal SqlServerTransportInfrastructure(QueueAddressTranslator addressTranslat
...
-            this.schemaAndCatalogSettings = settings.GetOrCreate<EndpointSchemaAndCatalogSettings>();
+            schemaAndCatalogSettings = settings.GetOrCreate<EndpointSchemaAndCatalogSettings>();
+            delayedDeliverySettings = settings.GetOrCreate<DelayedDeliverySettings>();
 
             //HINT: this flag indicates that user need to explicitly turn outbox in configuration.
             RequireOutboxConsent = true;
+
+            //Automatically disable the timeout manager support mode if it has been disabled via a feature toggle or endpoint is send-only (cannot run TM)
+            var timeoutManagerFeatureDisabled = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Disabled;
+            var sendOnlyEndpoint = settings.GetOrDefault<bool>("Endpoint.SendOnly");
```
bording: Will this native delayed delivery implementation work if `ConfigureReceiveInfrastructure` isn't called to start the `DelayedDeliveryMessagePump`? In other words, are all delayed messages kept in a local per-endpoint table even when they are for another endpoint? For RabbitMQ they are not, so there is no receiving dependency for deferring a message to another endpoint, allowing it to work for a send-only endpoint as well.

src/NServiceBus.SqlServer/DelayedDelivery/PreventRoutingMessagesToTimeoutManager.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace NServiceBus.Transport.SQLServer
+{
+    using Features;
+
+    class PreventRoutingMessagesToTimeoutManager : Feature
+    {
+        public PreventRoutingMessagesToTimeoutManager()
+        {
+            EnableByDefault();
```
bording: You've got the same `EnableByDefault` here that has been discovered to be problematic in RabbitMQ: https://github.com/Particular/NServiceBus.RabbitMQ/issues/394 https://github.com/Particular/NServiceBus.RabbitMQ/pull/393

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryInfrastructure.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    static class DelayedDeliveryInfrastructure
+    {
+        public static StartupCheckResult CheckForInvalidSettings(SettingsHolder settings)
+        {
+            var externalTimeoutManagerAddress = settings.GetOrDefault<string>("NServiceBus.ExternalTimeoutManagerAddress") != null;
+            if (externalTimeoutManagerAddress)
+            {
+                return StartupCheckResult.Failed("An external timeout manager address cannot be configured because the timeout manager is not being used for delayed delivery.");
+            }
```
bording: If you do need to remove the send-only setting from the `SqlServerTransportInfrastructure`, then you should add check to make sure `DisableTimeoutManager` and `SendOnly` haven't been called together at the same time, since that would lead to and invalid state.

SzymonPobiega:
@bording I've switched to opt-in mode which changed the behavior of the compat mode:
 * When send-only, you can't enable native delayed delivery (it requires receiving capability in SQL): https://github.com/Particular/NServiceBus.SqlServer/pull/364/files#diff-aa3c02805e020937f9c2ca3deaf71c7fR182
 * When native DD is enabled (which imples not-sendonly), there is a check to validate that external timeout manager has not been configured: https://github.com/Particular/NServiceBus.SqlServer/pull/364/files#diff-3cc262c7a1896d7009972765718d59d4R12
 * When user disabled `TimeoutManager` as a feature, it is honored the same way as if she/he did it via the explicit native DD APIs: https://github.com/Particular/NServiceBus.SqlServer/pull/364/files#diff-734321b72d52bc31733b86191d75e96cR28

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
```diff
@@ -171,6 +172,22 @@ public static TransportExtensions<SqlServerTransport> WithPeekDelay(this Transpo
         }
 
         /// <summary>
+        /// Enables native delayed delivery.
+        /// </summary>
+        public static DelayedDeliverySettings UseNativeDelayedDelivery(this TransportExtensions<SqlServerTransport> transportExtensions)
+        {
+            var sendOnlyEndpoint = transportExtensions.GetSettings().GetOrDefault<bool>("Endpoint.SendOnly");
+            if (sendOnlyEndpoint)
```
bording: This feels like the wrong place for this check. What happens If I call this first, and then make the endpoint send-only?

SzymonPobiega: Good catch! Fixed :-)

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
```diff
@@ -171,6 +172,22 @@ public static TransportExtensions<SqlServerTransport> WithPeekDelay(this Transpo
         }
 
         /// <summary>
+        /// Enables native delayed delivery.
+        /// </summary>
+        public static DelayedDeliverySettings UseNativeDelayedDelivery(this TransportExtensions<SqlServerTransport> transportExtensions)
+        {
```
weralabaj: that was mentioned before, since some of the settings are obligatory, wouldn't it make sense to specify them as parameters to this method? e.g. have signature like UseNativeDelayedDelivery(string tableSuffix, ...)

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryInfrastructure.cs
weralabaj: I don't understand this comment, @SzymonPobiega have you addressed it? ☝️

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageTable.cs
weralabaj: Should we remove this method?

src/NServiceBus.SqlServer/DelayedDelivery/MaturedDelayMessageHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Logging;
+
+    class MaturedDelayMessageHandler
+    {
+        public MaturedDelayMessageHandler(DelayedMessageTable table, SqlConnectionFactory connectionFactory, TimeSpan resolution, int batchSize)
```
weralabaj: > MaturedDelayMessageHandler

MaturedDelay **ed** MessageHandler?

src/NServiceBus.SqlServer/DelayedDelivery/MaturedDelayMessageHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            this.table = table;
+            this.connectionFactory = connectionFactory;
+            this.resolution = resolution;
```
weralabaj: > resolution 

I'd stick to calling it "interval" for consistency.

src/NServiceBus.SqlServer/DelayedDelivery/MaturedDelayMessageHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    using (var connection = await connectionFactory.OpenNewConnection().ConfigureAwait(false))
+                    {
+                        using (var transaction = connection.BeginTransaction())
+                        {
+                            await table.MoveMaturedMessages(batchSize, connection, transaction).ConfigureAwait(false);
```
weralabaj: Note to self: Seems like in the docs would be a good idea to stress that when timeout elapses messages are moved to the input queue, not dispatched immediately, so there's an additional reason why delivery time is approximate only. 

Question: BTW is there any reasoning we should present why it's this way rather than having additional, dedicated poller?

src/NServiceBus.SqlServer/Legacy/MultiInstance/LegacyMessageDispatcher.cs
```diff
@@ -1,25 +1,27 @@
...
     using System.Linq;
     using System.Threading.Tasks;
+    using System.Transactions;
     using Extensibility;
     using Transport;
 
     class LegacyMessageDispatcher : IDispatchMessages
     {
-        public LegacyMessageDispatcher(IQueueDispatcher dispatcher, LegacyQueueAddressTranslator addressTranslator)
+        public LegacyMessageDispatcher(LegacyQueueAddressTranslator addressTranslator, LegacySqlConnectionFactory connectionFactory)
```
weralabaj: Not clear to me why this change was necessary? Or was it just a refactoring?

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
SzymonPobiega: I think in case of SQL Server the table suffix should have a default (e.g. Delayed). We tend to use default names for things like this (e.g. satellites) and people are used to the fact that names that start with their endpoint name are reserved e.g. MyEndpoint.Timeouts.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryInfrastructure.cs
SzymonPobiega: Yeah. We don't allow native delayed delivery for send only  endpoints. Do our `DisableTimeoutManager` cannot be called for a send-only endpoint.

src/NServiceBus.SqlServer/DelayedDelivery/MaturedDelayMessageHandler.cs
SzymonPobiega: Nah, it is just us being "lazy" ;-) Seriously, we could have a dedicated poller but:
 * I think it might have negative impact on performance. The current version move a batch of messages to the input queue at a time and consuming the input queue is quite efficient. The alternative would be a consumer that operates based on the index on a due column which might be slow
 * A dedicated poller would have to implement all three transaction handling strategies making the code very complex

src/NServiceBus.SqlServer/Legacy/MultiInstance/LegacyMessageDispatcher.cs
SzymonPobiega: Just a refactoring, not related to this particular thing.

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
```diff
@@ -110,7 +152,43 @@ Expires ASC
...
+    FROM {1}.sys.objects
+    WHERE object_id = OBJECT_ID(N'{0}')
+        AND type in (N'U'))
+BEGIN
+    EXEC sp_releaseapplock @Resource = '{0}_lock'
+    RETURN
+END
+
+CREATE TABLE {0} (
+    Headers varchar(max) NOT NULL,
```
SzymonPobiega: Shouldn't we be consistent with the current code? I mean, it would not help us because the data from this table is moved to the input queue when the timeouts mature. So the funny characters would be lost anyway.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+                return DispatchBehavior.Deferred(doNotDeliverBefore.At, operation.Destination);
+            }
+            if (TryGetConstraint(operation, out delayDeliveryWith))
+            {
+                return DispatchBehavior.Deferred(DateTime.UtcNow + delayDeliveryWith.Delay, operation.Destination);
+            }
+            string expireString;
+            var headers = operation.Message.Headers;
+            if (headers.TryGetValue(TimeoutManagerHeaders.Expire, out expireString))
```
tmasternak: Why do we need to handle TimeoutManager messages here?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
```diff
@@ -0,0 +1,101 @@
...
+        static bool TryGetConstraint<T>(IOutgoingTransportOperation operation, out T constraint) where T : DeliveryConstraint
+        {
+            constraint = operation.DeliveryConstraints.OfType<T>().FirstOrDefault();
+            return constraint != null;
+        }
+
+        DelayedMessageTable delayedMessageTable;
+        ITableBasedQueueFactory immediateDeliveryQueueFactory;
+
+        struct DispatchBehavior
```
tmasternak: Do we need `DispatchBehavior`? It does need `Destination` as it's on the original message. Maybe `GetDueTime` could be `bool TryGetDueTime(out Timespan timeout)`

src/NServiceBus.SqlServer/DelayedDelivery/MaturedDelayedMessageHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Logging;
+
+    class MaturedDelayedMessageHandler
+    {
+        public MaturedDelayedMessageHandler(DelayedMessageTable table, SqlConnectionFactory connectionFactory, TimeSpan interval, int batchSize)
```
tmasternak: I'm not native speaker but `Matured` sounds off.  PS: I know that's not very constructive but can't think of anything better :/.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using DelayedDelivery;
+    using DeliveryConstraints;
+    using Performance.TimeToBeReceived;
+
+    class DelayedDeliveryTableBasedQueueFactory : ITableBasedQueueFactory
```
tmasternak: Name suggests that it's a factory for queues but I don't think it is.

tmasternak: I would like to question the decorator approach for handling `DelayedDelivery`. 

I think that delayed messages table is conceptually different thing than queue which happens to have similar schema to table based queue. I would suggest moving the delayed delivery logic to dispatcher. That would:
 * Make it explicit that native delays move message to-be-dispatched to a dedicated table
 * Change a bit confusing (at least for me) design in which each TableBasedQueue is figuring out if message is delayed or not

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
weralabaj: In case older endpoint still sends messages using TM.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
SzymonPobiega: @tmasternak we went with this design because of the opt-in character of the delayed delivery in 3.1.
Previously, when we wanted to release 4.0 and make native delayed delivery the only option the code was in the dispatcher . 
I would like to keep the decorator approach until native is the only option because it allows me to keep all the nasty `if`s in the infrastructure class.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
SzymonPobiega: Isn't it because of the SLR? We trick the core that we don't support delayed delivery so SLR messages contain the TM headers, not proper delivery constraints.

src/NServiceBus.SqlServer/DelayedDelivery/MaturedDelayedMessageHandler.cs
SzymonPobiega: @adamralph is matured a word? We would like to have a name that describes a delayed message which due time is in the past.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
SzymonPobiega: @tmasternak the destination might come from the operation or from a TM header (in case of SLRs)

src/NServiceBus.SqlServer/DelayedDelivery/MaturedDelayedMessageHandler.cs
adamralph: Are messages ever handled before they are due? I.e. can it not just be `DelayedMessageHandler`?

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
weralabaj: Any decisions from maintainers on this? Or will we wait for the unicode issue to go forward?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
weralabaj: I think this is a good argument, as it'll be easier to refactor code in the next version. Otherwise we'll have a lot of complex logic peppered in multiple places, so I also vote to keep as is.

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
SzymonPobiega: Yeha. The maintainers did make a decision to use `nvarchar` so we should adjust here.

weralabaj:
@MarcinHoppe @SzymonPobiega please see if I missed anything to change

SzymonPobiega:
@weralabaj I changed the param type for the command to be `NVarchar`. Now it should be all good. Question is: should I rebased it on the 'develop' effectively merging the multi-catalog and this PR? @MarcinHoppe ?

SzymonPobiega:
🎉

PR #372, opened 2017-06-27
Use explicit early-committed transaction in ReceiveOnly transaction mode

### LINKED ISSUE #368: In non-transactional mode the isolation level is inherited from the connection
### Who's affected

Users who run their endpoints in `TransportTransactionMode.None` mode.

### Symptoms

There is no explicit receive transaction so the isolation level is inherited from the connection. Because connections from the pool don't have their isolation level reset, if a user is using ADO.NET for data access in one of the non-supported isolation modes it can happen that SQL transports gets a connection in the unsupported isolation mode, causing receive to fail.

SzymonPobiega:
Fixes #368


SzymonPobiega:

PR #376, opened 2017-07-04
Support unicode headers

### LINKED ISSUE #340: Headers are stored in `varchar` column
### Symptoms

Characters not supported by database's collation are converted into `?` when used in headers (either as header name or value) leading to potential information loss.

### Who's affected

All users of SQL Server transport

### Description

The headers are stored in `varchar` column making in impossible to use higher Unicode symbols in the header name or value

SzymonPobiega:
Fixes #340 

 * Change the type of `Headers` column to `nvarchar(max)`
 * Change the parameter type of the send command to `NVarchar`
 * Add start-time validation (similar to `Expired` index check) that logs `WARN` if the `Headers` column type is `varchar` with CTA to change the type manually

### DIFF
--- a/src/NServiceBus.SqlServer/Receiving/SchemaVerification.cs
+++ b/src/NServiceBus.SqlServer/Receiving/SchemaVerification.cs
@@ -5,14 +5,14 @@
     using System.Threading.Tasks;
     using Logging;
 
-    class SchemaVerification
+    class SchemaInspector
     {
-        public SchemaVerification(Func<TableBasedQueue, Task<SqlConnection>> openConnection)
+        public SchemaInspector(Func<TableBasedQueue, Task<SqlConnection>> openConnection)
         {
             this.openConnection = openConnection;
         }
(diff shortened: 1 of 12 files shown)


SzymonPobiega:
@Particular/sqlserver-transport-maintainers this is now ready for the review.

src/NServiceBus.SqlServer/Receiving/SchemaVerification.cs
```diff
@@ -0,0 +1,62 @@
+﻿namespace NServiceBus.Transport.SQLServer
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Threading.Tasks;
+    using Logging;
+
+    class SchemaVerification
```
MarcinHoppe: I somehow think this class could be named better. It is named after an action (verification) rather than after an actor that executes this action (verifier). Maybe `SchemaInspector` would be better?

PR #378, opened 2017-07-07
Index_Expires should be filtered

### LINKED ISSUE #323: Evaluate filtered index on Expires column
### What is the problem?

Currently an index on `Expires` column indexes all the rows. If an endpoints exchange messages without the `TimeToBeReceived` (TTBR) attribute set, the index will mostly be useless but we will pay the cost of maintaining it with each send/receive/purge operation.

We could use a filtered index instead and use `Expires IS NOT NULL` condition to only index rows with TTBR.

### What to do about it?

Change index definition to a filtered index and evaluate performance of:

1. Send with TTBR,
1. Peek with TTBR,
1. Receive with TTBR,
1. Purge expired messages with some TTBR messages in the queue,
1. Send without TTBR,
1. Peek without TTBR,
1. Receive without TTBR,
1. Purge expired messages with no TTBR messages in the queue.

MarcinHoppe:
Resolves #323

### DIFF
--- a/src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
+++ b/src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
@@ -152,6 +152,8 @@ INCLUDE
     Id,
     RowVersion
 )
+WHERE
+    Expires IS NOT NULL
 
 EXEC sp_releaseapplock @Resource = '{0}_lock'";
 


SzymonPobiega:
Should we add this to 3.1? Do we need to add any information for the users to encourage them to re-run the scripts to add this filtering?

MarcinHoppe:
It's a small change, I don't think we need to add this to 3.1 or encourage people to re-create this index.

SzymonPobiega:
@MarcinHoppe but it wouldn't be bad if we pulled it to 3.1? I need to generate the final build anyway...

SzymonPobiega:
BTW, do we have 2 review policy in SQL? Meaning I need to wait for @tmasternak to approve it?

MarcinHoppe:
No, it would not hurt :). My intent was not to delay 3.1 even further.

MarcinHoppe:
> BTW, do we have 2 review policy in SQL?

All replicas must reach consensus.

tmasternak:
LGTM

SzymonPobiega:
Consensus reached in Epoch 0.

PR #379, opened 2017-07-12
Data corruption when using Unicode characters in message headers on SQL Transport

MarcinHoppe:
Replaced by #381 (wrong branch name)


MarcinHoppe:
Wrong branch name, closing.

PR #380, opened 2017-07-13
Data corruption when using Unicode characters in message headers on SQL Transport

MarcinHoppe:
Replaced by #382 (wrong branch name).


MarcinHoppe:
Wrong branch name, closing. Replaced by #382.

PR #383, opened 2017-07-31
Usewildcardsfortests

SimonCropp:
allow tests to auto update and run on the correct version without needing to do a nuget update


SimonCropp:
failing test seems unrelated

PR #384, opened 2017-07-31
Cleanup

SimonCropp:


src/NServiceBus.SqlServer/Configuration/ConnectionPoolValidator.cs
```diff
@@ -15,15 +15,14 @@ public static ValidationCheckResult Validate(string connectionString)
                 return ValidationCheckResult.Valid();
             }
 
-            if (!keys.ContainsKey("Max Pool Size"))
+            if (keys.ContainsKey("Max Pool Size"))
```
MarcinHoppe: I'm not sure what is the value of this change is, but I'll be happy to learn.

SimonCropp: two things. 

 1. the fewer logic operations in code the easier it is to reason about. when reading/reviewing the code and u see a negation that is a mental operation your brain needs to store. when reviewing the next few lines. it is why people often get confused by double negatives.
 2. a single negation `!` is easy to miss and stuff up the logic

so when writing code, if the inversion of the negation is semantically equivalent, i will always avoid the negation.

MarcinHoppe: I can agree for a single conditional. Already merged.

SimonCropp: yeah it is not always so clear. there are also other things like reducing nesting and early exits to consider.

PR #385, opened 2017-08-01
update fody

SimonCropp:
included fixes that will be required to support .net standard


SimonCropp:
@Particular/sqlserver-transport-maintainers this one is ready to go

PR #390, opened 2017-08-09
Removes references to NHibernate persistence

SzymonPobiega:
In order to not block V7-compatible release.


MarcinHoppe:
How are we going to verify that we're not escalating in the future without this test?

SimonCropp:
@MarcinHoppe 

> How are we going to verify that we're not escalating in the future without this test?

i think we pull that into a diff repo+sln the same way we have the wire compat tests?

MarcinHoppe:
Wire compat tests recently travelled exactly the opposite route: #353 and are now hosted in this repo.

SzymonPobiega:
I think that breaking this so unlikely that a proper sample would be enough. We do smoke testing via samples before releases anyway.

MarcinHoppe:
I beg to disagree with this:

> I think that breaking this so unlikely

The code around transactions, especially in the transports, was quite tricky to write. I'd prefer to have the safeguard of an automated test to cover this non-trivial guarantee that we give to people using SQL Server only.

SzymonPobiega:
So the alternative would be to add references to NH and SQL persistences in the compat test facades. This way we can have a compat tests that verifies it. Since facades are version-specific, we will just skip the NH in the SQL 4 facade.

MarcinHoppe:
👍

SimonCropp:
@MarcinHoppe 

> Wire compat tests recently travelled exactly the opposite route: #353 and are now hosted in this repo

yes and if you remember i argued against that because it would cause problems long term

MarcinHoppe:
I recall there were arguments both ways. So far I have not heard it is causing problems so we might give this direct a benefit of a doubt and see if it causes pain down the road.

@Particular/sqlserver-transport-maintainers Thoughts?

SzymonPobiega:
I've added two integration tests projects for SQL and NH persistences that verify the escalation. When we switch to V7 with SQL transport develop then we'll remove the NH tests untill NH has a V7 compatible release. Does this sound OK?

MarcinHoppe:
Sounds like a workable plan. I'll merge when the build is green.

PR #399, opened 2017-09-08
Do not expose GitVersionTask as a dependency

### LINKED ISSUE #398: Why does this package depend on GitVersionTask?
Why does this package depend on GitVersionTask? I can understand if this is a development dependency, but it seems wrong that it is a production dependency.

MarcinHoppe:
Resolve #398


MarcinHoppe:
@Particular/sqlserver-transport-maintainers Ready to merge and release the new beta.

PR #400, opened 2017-09-26
Run APIApprovals for correct assembly

bording:
While reviewing https://github.com/Particular/NServiceBus/pull/4971, I noticed that SqlServer had a reference to the `RijndaelKeyIdentifier` that it shouldn't have.

Looks like when the APIApprovals test was updated, it was still using a core type, so the wrong assembly was being compared.


bording:
The test failure here is unrelated to this change. Looks like there are some issues with the project files. I'm going to submit a new PR to fix those up.

bording:
Noticed that there's a release branch that has this fixed already. One more reason why I'm 👎 for release branches for betas.

I'll look and see if there's any additional project cleanup that needs to be done still on the release branch.

Issue #396, opened 2017-09-04
Purge expired messages at startup (and not periodically)


MarcinHoppe:
### What is the problem?

Users who do not use Time-To-Be-Received (TTBR) with the SQL Server transport still run the expired message purge task that unnecessarily increases the load on the queue table.

### What to do about it?

Revert to the old behaviour where the message pump discards expired messages when they are received. Purge the expired messages at startup only to support endpoints that cannot start because of high number of expired messages in the incoming queue table.

SzymonPobiega:
@MarcinHoppe how about we run that task only when we detect a TTBR message? Or at least 1 message per some time unit? This way users would not have another knob to thing about. Say if there is at least one expired TTBRed message per second detected, run the cleaner.

MarcinHoppe:
Magic is nice but for troubleshooting, explicit control is nicer. This issue was created from a support case.

I will not object too much either way.

SzymonPobiega:
@MarcinHoppe we can have both. Magic will ensure it consumes less resources for the ignorant majority (people who don't care what TTBR even is) and the explicit "do not run cleanup" will comfort the ones that are aware of the trade offs.

MarcinHoppe:
That I can live with 😄.

SzymonPobiega:
Looking at the code in 3.x we don't receive the timed-out messages so we can't just disable the purge automatically because we would not get feedback if it is really needed.  I don't think it makes sense to have a sophisticated policy for how often to run the purge since it would be fragile.

Now that I think about it, what is the additional load if the expired message purge query finds no records to delete? Is it worth doing?

On the other hand, the 2.2.x version which is still used by SC has a number of flows that we did not address:
 * It uses unbounded deletes in the expired purge
 * The receive overlaps with the expired purge because it does not exclude the expired messages
 * The purge uses `UPDLOCK` instead of `NOLOCK`.

SzymonPobiega:
@MarcinHoppe ping?

MarcinHoppe:
@SzymonPobiega How about:

```csharp
var transport = endpointConfiguration.UseTransport<SqlServer>();
transport.DoNotPurgeExpiredMessages();
```

?

SzymonPobiega:
Ok, I was mistaken twice actually and the original thing I wrote is true:

[...]

PR #406, opened 2017-10-19
Expired message are only purged on startup and only if explicitly enabled

SzymonPobiega:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/396

I decided to require explicit enabling of the purge-on-startup feature because in a scale-out environment purging on startup by default can cause problems.

### DIFF
--- a/src/NServiceBus.SqlServer.sln
+++ b/src/NServiceBus.SqlServer.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26730.3
+VisualStudioVersion = 15.0.26730.12
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NServiceBus.SqlServer", "NServiceBus.SqlServer\NServiceBus.SqlServer.csproj", "{FA1193BF-325C-4201-BB78-484032E09809}"
 EndProject
@@ -35,6 +35,8 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Facade_3_1", "CompatibilityTests\Facade_3.1\Facade_3_1.csproj", "{66D10F43-3859-4F55-B8A8-60E32AD6B813}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NServiceBus.SqlServer.IntegrationTests.Sql", "NServiceBus.SqlServer.IntegrationTests.Sql\NServiceBus.SqlServer.IntegrationTests.Sql.csproj", "{15430E85-0954-445B-9F1D-4E5D321AE9BB}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NServiceBus.SqlServer.TestHarness", "NServiceBus.SqlServer.TestHarness\NServiceBus.SqlServer.TestHarness.csproj", "{373C5E59-6994-4120-A569-8EAA8F4A8CC7}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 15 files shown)


src/NServiceBus.SqlServer/Receiving/MessagePump.cs
```diff
@@ -89,6 +87,7 @@ public void Start(PushRuntimeSettings limitations)
 
         async Task ProcessMessages()
         {
+            await PurgeExpiredMessages().ConfigureAwait(false);
```
MarcinHoppe: I think it would be more readable to call this right next to to the normal purger in `MessagePump.Init`. Logically, purging expired messages does not seem like a part of message processing.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
```diff
@@ -8,22 +8,22 @@
...
         }
 
-        public TimeSpan PurgeTaskDelay { get; }
-
-        int PurgeBatchSize { get; }
-
         public async Task Purge(TableBasedQueue queue, CancellationToken cancellationToken)
         {
-            Logger.DebugFormat("Starting a new expired message purge task for table {0}.", queue);
+            if (!enable)
```
MarcinHoppe: Can we move this enable / disable logic elsewhere? One possible option is to return null from `SqlServerTransportInfrastructure.CreateExpiredMessagesPurger` when `enable` is false and then use the Elvis operator when calling it:

```csharp
await expiredMessagesPurger?.Purge(inputQueue, cancellationToken).ConfigureAwait(false);
```

or even short-circuit before that.

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
```diff
@@ -48,14 +48,22 @@ public void PrepareSendCommand(SqlCommand command)
         static async Task<MessageRow> ReadRow(SqlDataReader dataReader)
         {
             //HINT: we are assuming that dataReader is sequential. Order or reads is important !
+            var id = await dataReader.GetFieldValueAsync<Guid>(0).ConfigureAwait(false);
```
MarcinHoppe: Why create all those variables?

src/NServiceBus.SqlServer.sln
```diff
@@ -36,6 +36,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Facade_3_1", "Compatibility
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NServiceBus.SqlServer.IntegrationTests.Sql", "NServiceBus.SqlServer.IntegrationTests.Sql\NServiceBus.SqlServer.IntegrationTests.Sql.csproj", "{15430E85-0954-445B-9F1D-4E5D321AE9BB}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NServiceBus.SqlServer.TestHarness", "NServiceBus.SqlServer.TestHarness\NServiceBus.SqlServer.TestHarness.csproj", "{373C5E59-6994-4120-A569-8EAA8F4A8CC7}"
```
MarcinHoppe: Was adding this test harness intentional?

SzymonPobiega: I was on the fence regarding that. It was not the first time I created something like this in the solution to be able to easily see the impact of my changes. What do you think?

src/NServiceBus.SqlServer/Queuing/MessageRow.cs
SzymonPobiega: I'll revert to the previous state. That's a relic of my experiments

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
SzymonPobiega: I don't like passing nulls to the pump. The pump does not care if the purging is enabled or not, it just cares about invoking it before starting to process messages.

I thought about creating a `NullPurger` thing that implements the same interface as an empty method but figured this would be too much ceremony. Having a switch in the purger seems like a middle ground.

src/NServiceBus.SqlServer/Receiving/MessagePump.cs
SzymonPobiega: 👍

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
MarcinHoppe: Let's keep it the way it is now, looks like a decent trade-off.

src/NServiceBus.SqlServer.sln
MarcinHoppe: I have an idea: we can move the test harness to a separate subfolder with its own solution file that includes only the transport project and the test harness. This way it's easy to work with (no test projects) and does not clutter the main solution file.

src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs
SzymonPobiega: 👍

PR #410, opened 2017-11-06
Replaces the usages of Settings.LocalAddress and Settings.LogicalAddress

SzymonPobiega:
In order to be able to run SQL Server transport in NSB.Raw


MarcinHoppe:
Is this required for the ServiceControl transport adapter?

SzymonPobiega:
@MarcinHoppe nope, only for NSB.Raw and if, for some reason, somebody configures SC.TA to enable native delayed delivery

PR #412, opened 2017-11-09
[WIP] Retry when queue table creation deadlocks

MarcinHoppe:


MarcinHoppe:
The experiment failed, the deadlocks are still there.

PR #413, opened 2017-11-10
Explicitly set size for Body parameter when inserting in queue

jmezach:
After moving our production environment to NServiceBus 6 with SQL transport we noticed a very high number of execution plans for the insert queries on the various queue tables. Further analysis revealed that this was caused by the Body parameter being passed as a VARBINARY(<size of byte array>), which means that for every different size of message SQL Server will generate a new execution plan.

By setting the Size property on the parameter to -1, the parameter is passed to SQL Server as a VARBINARY(MAX). That means that every execution of the query can use the same execution plan.

### DIFF
--- a/src/NServiceBus.SqlServer/Queuing/MessageRow.cs
+++ b/src/NServiceBus.SqlServer/Queuing/MessageRow.cs
@@ -42,7 +42,7 @@
             AddParameter(command, "Recoverable", SqlDbType.Bit, recoverable);
             AddParameter(command, "TimeToBeReceivedMs", SqlDbType.Int, timeToBeReceived);
             AddParameter(command, "Headers", SqlDbType.NVarChar, headers);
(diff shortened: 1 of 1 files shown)


MarcinHoppe:
@jmezach The 3.1.3 hotfix should go off the `master` branch. I took the liberty of changing the base branch.

jmezach:
@MarcinHoppe Ah, good catch. I checked out the tag for the previous release and build my fix on top of that. I see that the compile is failing now, but I cannot see the reason why. Any updates on this?

MarcinHoppe:
@jmezach Our internal GitVersion gurus suggested that you change the branch name to follow the `hotfix-major.minor.patch` format, i.e. to `hotfix-3.1.3`. Can you rename the branch and submit a new PR to `master`?

jmezach:
Done. See #415. Seems to fail as well though.

PR #414, opened 2017-11-15
[WIP] Add a test showing providing external connection for a send-only endpoint

SzymonPobiega:


src/NServiceBus.SqlServer.TransportTests/NServiceBus.SqlServer.TransportTests.csproj
```diff
@@ -16,4 +16,8 @@
     <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
```
MarcinHoppe: What's this?

src/NServiceBus.SqlServer.AcceptanceTests/When_providing_external_connection.cs
```diff
@@ -0,0 +1,87 @@
...
+                        transportTransaction.Set(transaction);
+
+                        var options = new SendOptions();
+                        options.GetExtensions().Set(transportTransaction);
+                        await bus.Send(new Message(), options);
+
+                        transaction.Commit();
+                    }
+                }))
+                .Done(c => c.MessageReceived)
```
MarcinHoppe: How do we know if the transport used the provided connection and transaction? Do we need more assertions here? Maybe rolling back the transaction after `bus.Send` would be a better indication that the same transaction is used?

SzymonPobiega: True. I forgot that the connection string is set by the test infrastructure.

src/NServiceBus.SqlServer.TransportTests/NServiceBus.SqlServer.TransportTests.csproj
SzymonPobiega: VisualStudio adds this crap automatically. I'll remove.

PR #415, opened 2017-11-16
Explicitly set size for Body parameter when inserting in queue

jmezach:
After moving our production environment to NServiceBus 6 with SQL transport we noticed a very high number of execution plans for the insert queries on the various queue tables. Further analysis revealed that this was caused by the Body parameter being passed as a VARBINARY(<size of byte array>), which means that for every different size of message SQL Server will generate a new execution plan.

By setting the Size property on the parameter to -1, the parameter is passed to SQL Server as a VARBINARY(MAX). That means that every execution of the query can use the same execution plan.

Replaces #413

### DIFF
--- a/src/NServiceBus.SqlServer/Queuing/MessageRow.cs
+++ b/src/NServiceBus.SqlServer/Queuing/MessageRow.cs
@@ -42,7 +42,7 @@
             AddParameter(command, "Recoverable", SqlDbType.Bit, recoverable);
             AddParameter(command, "TimeToBeReceivedMs", SqlDbType.Int, timeToBeReceived);
             AddParameter(command, "Headers", SqlDbType.NVarChar, headers);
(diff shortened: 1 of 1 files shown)


MarcinHoppe:
@jmezach Can you rebase on top of master? It seems the PR was created against develop again.

jmezach:
@MarcinHoppe I tried rebasing, but it says it's already up to date

MarcinHoppe:
@jmezach I think I know what is going on. When TeamCity clones the repo for building, it does not have a remote pointing at your repo (I do, however). Under the hood, GitVersion has to create a fake branch for your commit and since that fake branch does not have a version number in it, it does its usual magic and determines that it is 4.1.

I will close this PR and cherry-pick your commit into a new hotfix branch which does not come from a non-`origin` remote. Stay tuned.

PR #423, opened 2018-01-15
`readpast` when purging expired messages

SzymonPobiega:


tmasternak:
@SzymonPobiega I have removed the `nolock` as it otherwise overrides the effect of `readpast` hint.

PR #425, opened 2018-01-22
add missing sql connection to TransportTransaction

SimonCropp:
The current connection is exposed in these two places

  * https://github.com/Particular/NServiceBus.SqlServer/blob/release-4.0/src/NServiceBus.SqlServer/Receiving/ProcessWithNativeTransaction.cs#L66
 * https://github.com/Particular/NServiceBus.SqlServer/blob/release-4.0/src/NServiceBus.SqlServer/Receiving/ProcessWithNoTransaction.cs#L32

but not here https://github.com/Particular/NServiceBus.SqlServer/blob/release-4.0/src/NServiceBus.SqlServer/Receiving/ProcessWithTransactionScope.cs#L63

This adds the connection to the 3rd case as well

### DIFF
--- a/src/NServiceBus.SqlServer/Receiving/ProcessWithTransactionScope.cs
+++ b/src/NServiceBus.SqlServer/Receiving/ProcessWithTransactionScope.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Transport.SQLServer
 {
     using System;
+    using System.Data.SqlClient;
     using System.Threading;
     using System.Threading.Tasks;
     using System.Transactions;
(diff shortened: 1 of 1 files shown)


src/NServiceBus.SqlServer/Receiving/ProcessWithTransactionScope.cs
```diff
@@ -35,7 +36,7 @@ public ProcessWithTransactionScope(TransactionOptions transactionOptions, SqlCon
 
                     connection.Close();
```
SzymonPobiega: @SimonCropp the connection is not exposed in this mode because it is immediately closed, even before invoking the pipeline. The reason for it is that in transaction mode SQL Server can use execute multiple connections in a single local transaction as long as the connections don't overlap.

SzymonPobiega:
@SimonCropp I've left my comment inline.

SimonCropp:
@SzymonPobiega thanks for the clarification

PR #428, opened 2018-02-14
Add startup check for ambient transaction enlistment support

bording:
This ensures that the transport will fail on startup if the `TransactionScope` transport transaction mode has been selected when running against versions of the System.Data.SqlClient package that don't support ambient transaction enlistment.

The first version of the package that will support this is the 4.5.0 release, which is part of .NET Core 2.1.

FYI @timbussmann @andreasohlund @tmasternak

### DIFF
--- a/src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
+++ b/src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
@@ -82,7 +82,7 @@ namespace NServiceBus.Transport.SQLServer
             return new TransportReceiveInfrastructure(
                 () =>
                 {
-                    var pump = new MessagePump(receiveStrategyFactory, queueFactory, queuePurger, expiredMessagesPurger, queuePeeker, schemaVerification,  waitTimeCircuitBreaker);
+                    var pump = new MessagePump(receiveStrategyFactory, queueFactory, queuePurger, expiredMessagesPurger, queuePeeker, schemaVerification, waitTimeCircuitBreaker);
                     if (delayedDeliverySettings == null)
                     {
                         return pump;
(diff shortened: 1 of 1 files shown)


src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -141,6 +141,33 @@ ExpiredMessagesPurger CreateExpiredMessagesPurger(SqlConnectionFactory connectio
...
+                {
+                    using (var scope = new TransactionScope(TransactionScopeOption.RequiresNew, transactionOptions, TransactionScopeAsyncFlowOption.Enabled))
+                    using (var connection = await connectionFactory.OpenNewConnection().ConfigureAwait(false))
+                    {
+                        scope.Complete();
+                    }
+                }
+                catch (NotSupportedException)
+                {
+                    var message = "The version of System.Data.SqlClient in use does not support enlisting SQL connections in ambient transactions, so the TransactionScope transport transaction mode cannot be used. Use `EndpointConfiguration.UseTransport<SqlServerTransport>().Transactions` to select a different transport transaction mode.";
```
andreasohlund: Can we detect if the mode was defaulted by us or requested by the user and adjust the message accordingly?

bording: I'm not sure what sort of different message you'd want to see there. What do you have in mind?

andreasohlund:
Thinking that if they explicitly asked for TXScope it would be more likely
that they wanted DTC but lets call YAGNI on that

timbussmann:
tested this PR against the SQL sample and the behavior seems to work fine. One question though: This exception is also thrown on send-only endpoints, which actually work fine with the transactionscope tx mode, as they never open a receive transaction. So theoretically we don't have to enforce a configuration downgrade on send-only endpoints.

PR #433, opened 2018-03-29
add computed column to convert body to nvarchar

SimonCropp:
makes it easier to debug when looking at the raw sql tables
image

### DIFF
--- a/src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
+++ b/src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
@@ -139,6 +139,7 @@ CREATE TABLE {0} (
     Expires datetime,
     Headers nvarchar(max) NOT NULL,
     Body varbinary(max),
+    BodyString as cast(Body as nvarchar(max)),
     RowVersion bigint IDENTITY(1,1) NOT NULL
 );
 
@@ -184,6 +185,7 @@ END
 CREATE TABLE {0} (
     Headers nvarchar(max) NOT NULL,
     Body varbinary(max),
+    BodyString as cast(Body AS nvarchar(max)),
     Due datetime NOT NULL,
     RowVersion bigint IDENTITY(1,1) NOT NULL
 );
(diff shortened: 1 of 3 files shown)


src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
```diff
@@ -139,6 +139,7 @@ ELSE 1
     Expires datetime,
     Headers nvarchar(max) NOT NULL,
     Body varbinary(max),
+    BodyString as cast(Body as nvarchar(max)),
```
tmasternak: Do we know what is performance cost of adding that column? It's not used in any other query so I would assume it's calculated only when needed. That being said it would be nice to confirm this.

SimonCropp:
yep as it is non-persisted there is no perf impact, unless people explcitly query it

SimonCropp:
bump. need anything else from me?

SzymonPobiega:
What happens is somebody uses a binary serializer?

SimonCropp:
@SzymonPobiega dont know

SzymonPobiega:
@SimonCropp would adding a doco like this be enough https://docs.particular.net/transports/sql/operations-scripting ? We're concerned about adding complexity to the transport itself.

SimonCropp:
@SzymonPobiega does this proposal add significant complexity?

SzymonPobiega:
@SimonCropp well, it adds complexity to the scripts for the purpose of debugging plus it does not apply to binary serializers (for which the computed column contains junk characters). If someone is using SSMS for troubleshooting she/he has to query the table anyway. Copying the query script from the doco seems like a good option, especially that it is tailored for this job (omits unused columns and locked records).

SimonCropp:
given the majority of users use a text based serializer (xml or json) i still think this deserves more than just "place it in doco". I think that would result the potential value being very rarely realised. can we make it an optional config option? i would prefer it was "on by default" but an "opt in" setting would be sufficient.

SzymonPobiega:
@tmasternak can you chime in? I am on the fence here

SzymonPobiega:
OK. Makes sense. You convince me.

SzymonPobiega:
@SimonCropp Merging it now. It will be released in 4.1. We don't want to touch 4.0 at this stage. Thanks :heart:

SimonCropp:
@SzymonPobiega playing around with this a bit more i think the computed column sql text needs to be a configurable option. depending on encoding and byte order marks 

image

SzymonPobiega:
@SimonCropp :+1: . I've added it to the next release https://github.com/Particular/NServiceBus.SqlServer/issues/445

PR #439, opened 2018-05-11
[WIP] Prepare RTM

bording:
Marking this WIP for now because it's still referencing an RC version of NServiceBus.Persistence.Sql in the NServiceBus.SqlServer.IntegrationTests.Sql project.

We need to resolve the circular reference between the repos!


bording:
The circular reference has been resolved by removing the reference to NServiceBus.Persistence.Sql, so this PR is ready to go now!

PR #440, opened 2018-05-11
Remove persistence tests to avoid circular reference between repos

bording:
This removes the NServiceBus.SqlServer.IntegrationTests.Sql project. The test in that project has been moved to https://github.com/Particular/NServiceBus.Persistence.Sql/pull/279.

I noticed that there was also a NServiceBus.SqlServer.IntegrationTests.NHibernate project, but it was not being referenced in any of the repo's solution files.

Should I move this test to the NServiceBus.NHibernate repo as well?


bording:
FYI I've already fixed the inspection failure in #439, so it will be okay to merge this with the failure and it will be fixed over there.

SzymonPobiega:
@bording the NH project has been removed at some point because there was no stable version on NHibernate, if my memory is not failing me. I like this change. Anybody against it?

PR #443, opened 2018-05-16
Don't flood logs

SzymonPobiega:
I decided to use two values instead of a classic exponential back-off because:
 * We don't ever want to wait for less than 1 second
 * We don't ever want to wait for more than 10 seconds so that when the DB is back online we can quickly resume processing

Based on these bounds with exponential back-off we would only permit `c` values of 1,2 and 3 (2^3=8) which gives little benefit for the added complexity.

In case when `peek` works fine but `receive` fails (i.e. lack of `DELETE` permissions) the receive circuit breaker is going to be triggered. To prevent concurrently hammering the database the message pump has been modified to run only a single receive task at a time if the receive CB is triggered.

### DIFF
--- a/src/NServiceBus.SqlServer/Receiving/RepeatedFailuresOvertimeCurcuitBreaker.cs
+++ b/src/NServiceBus.SqlServer/Receiving/RepeatedFailuresOvertimeCurcuitBreaker.cs
@@ -16,7 +16,7 @@
             timer = new Timer(CircuitBreakerTriggered);
         }
 
-        public bool Triggered { get; private set; }
+        public bool Triggered => triggered;
 
         public void Dispose()
         {
@@ -33,7 +33,7 @@
             }
 
             timer.Change(Timeout.Infinite, Timeout.Infinite);
-            Triggered = false;
+            triggered = false;
             Logger.InfoFormat("The circuit breaker for {0} is now disarmed", name);
         }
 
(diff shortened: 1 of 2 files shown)


src/NServiceBus.SqlServer/Receiving/RepeatedFailuresOvertimeCurcuitBreaker.cs
```diff
@@ -45,18 +48,21 @@ public Task Failure(Exception exception)
...
-            return Task.Delay(TimeSpan.FromSeconds(1));
+            var delay = Triggered ? ThrottledDelay : NonThrottledDelay;
+            return Task.Delay(delay);
         }
 
         void CircuitBreakerTriggered(object state)
         {
             if (Interlocked.Read(ref failureCount) > 0)
             {
+                Triggered = true;
```
Scooletz: This should be either a `volatile` field or a field that is written with `Volatile.Write` and read with `Volatile.Read`. Otherwise, the change applied in here might not become visible to the reader in line 51. Also, we never set the `Triggered` to false so once triggered, forever triggered it is.

Scooletz: Oh my. Github collapsed it with `Dispose` and I didn't notice it.

SzymonPobiega:
Something is off with the tests. Waiting for a green build.

Issue #445, opened 2018-05-29
Make adding a computed column for viewing message body an opt-in


SzymonPobiega:
The PR adding a computed column has been merged to develop, however, it should be a configurable opt-in. The goal is to add methods that allow to turn it on, it will be disabled by default.

- [x] Add info in doco (the feature hasn't been documented yet) https://github.com/Particular/docs.particular.net/pull/3861/commits/e1ee79a97436458fc956a81e1cd7df79b5112c1d

Based on [this comment](https://github.com/Particular/NServiceBus.SqlServer/pull/433#issuecomment-384830631) we might need to make this feature configurable to account for BOM or different encodings.

![image](https://user-images.githubusercontent.com/122666/39338794-736a1c2e-4a08-11e8-9e13-338054f16957.png)

SimonCropp:
so in sql native i allowed people to override it https://github.com/NServiceBusExtensions/NServiceBus.Native/blob/288e507e930a228e0d858d6d722b1a13816d5d7e/SqlServer.Native/BaseQ/BaseQueueManager_QueueCreator.cs#L34

PR #446, opened 2018-06-12
Message body computed column

weralabaj:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/445

### DIFF
--- a/src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryQueueCreator.cs
+++ b/src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryQueueCreator.cs
@@ -40,7 +40,9 @@ namespace NServiceBus.Transport.SQLServer
             }
             if (createMessageBodyComputedColumn)
             {
+#pragma warning disable 618
                 var bodyStringSql = string.Format(SqlConstants.AddMessageBodyStringColumn, canonicalQueueAddress.QualifiedTableName, canonicalQueueAddress.QuotedCatalogName);
+#pragma warning restore 618
                 using (var command = new SqlCommand(bodyStringSql, connection, transaction)
                 {
                     CommandType = CommandType.Text
--- a/src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryQueueCreator.cs
+++ b/src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryQueueCreator.cs
@@ -21,15 +21,15 @@ namespace NServiceBus.Transport.SQLServer
             using (var transaction = connection.BeginTransaction())
             {
                 await CreateDelayedMessageQueue(delayedMessageTable, connection, transaction, createMessageBodyComputedColumn).ConfigureAwait(false);
+
                 transaction.Commit();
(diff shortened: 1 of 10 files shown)


weralabaj:
FYI, reviewers, I'm looking into the failed test.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryQueueCreator.cs
```diff
@@ -41,5 +43,6 @@ public DelayedDeliveryQueueCreator(SqlConnectionFactory connectionFactory, ICrea
         SqlConnectionFactory connectionFactory;
         ICreateQueues queueCreator;
         CanonicalQueueAddress delayedMessageTable;
+        readonly bool createMessageBodyComputedColumn;
```
SzymonPobiega: no `readonly`

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
```diff
@@ -137,7 +139,7 @@ ELSE 1
     Recoverable bit NOT NULL,
     Expires datetime,
     Headers nvarchar(max) NOT NULL,
-    BodyString as cast(Body as nvarchar(max)),
+    {2}
```
SzymonPobiega: How about adding this computed column as a separate `ALTER TABLE` statement executed right after create?

weralabaj: Hm... at first I jumped at this suggestion because I thought it will be more elegant, especially if we move the call a bit up the call stack we could avoid passing the argument down to all methods so many times, e.g. by moving the alter table command up to the `CreateQueueIfNecessary` method. 

However, in that case, I also have to either add the alter command to the create queue text (in that same place in fact where I format the query string at the moment) and execute them together or make one more separate call to the database.

So... I'm not convinced. Am I missing something?

SzymonPobiega: > execute them together or make one more separate call to the database.

I am 👍 for executing them as separate commands withing the same transaction

weralabaj: Hm... not my personal preference, but you know better :D

SzymonPobiega: LoL. I wanted to minimize the amount of string.formatting that happens on these commands. I already hate that the table name can't be parametrized.

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
```diff
@@ -118,6 +118,14 @@ ELSE 1
         AND type in (N'U'))
 RETURN
 
+IF EXISTS (
```
SzymonPobiega: I think we should do this inside the lock to prevent two instances on an endpoint creating these columns concurrently.

weralabaj: I didn't consider this scenario! Thanks

Issue #312, opened 2016-10-07
Issue warning regarding isolation level settings on the SqlTransport


tmasternak:
## PoA

- [x] Issue a warning if isolation level is below readcommited https://github.com/Particular/NServiceBus.SqlServer/issues/312
- [x] Update doco https://github.com/Particular/docs.particular.net/pull/3861/commits/3d40b3080a87f4150117671730f7695557a673dc
- [x] Raise an issue to provide a fix in plat dev https://github.com/Particular/PlatformDevelopment/issues/2101

## Context

Right now we enable users to set transaction isolation levels that will be used for receive operations and for send operations (but not in all cases e.g. running with Outbox, Isolated Dispatches, ReceiveOnly and None transactions modes).

I think it does not make sense to run the transport in any other IsolationLevel than ReadCommitted or RepeatableRead. 
- For lower levels + Snapshot we can have message processed twice - we do not have locking on row
- For Serializable - we would have to rewrite queries as we use hints available only there + we would probably not gain anything in terms of processing logic

I propose to change api and enable transaction timeout settings and enforce ReadCommitted for all operations.

See: #310 for isolation level setup details.

/cc: @Particular/sqlserver-transport-maintainers @ramonsmits

ramonsmits:
Yes, limit the transaction isolation level for transport receive operations. It doesn't make sense.

PR #447, opened 2018-06-12
add a warning

weralabaj:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/312

### DIFF
--- a/src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
+++ b/src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
@@ -6,6 +6,7 @@
     using System.Transactions;
     using Configuration.AdvancedExtensibility;
     using Features;
+    using Logging;
 
     /// <summary>
     /// Adds extra configuration for the Sql Server transport.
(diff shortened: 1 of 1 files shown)


src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
```diff
@@ -151,6 +152,11 @@ public static TransportExtensions<SqlServerTransport> TransactionScopeOptions(th
         {
             Guard.AgainstNull(nameof(transportExtensions), transportExtensions);
 
+            if (isolationLevel != IsolationLevel.ReadCommitted)
```
SzymonPobiega: Here's the doco page about the supported levels: https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-2017. We are interested mostly in the `readpast` hint as it is required to efficiently delete rows. The doco says:
> READPAST can only be specified in transactions operating at the READ COMMITTED or REPEATABLE READ isolation levels. When specified in transactions operating at the SNAPSHOT isolation level, READPAST must be combined with other table hints that require locks, such as UPDLOCK and HOLDLOCK.

And then elaborates on more ezoteric scenarios that involve snapshot isolation. I think we can skip the ezoteric stuff in this release but I think we should include both ReadCommitted and RepeatableRead as supported values.

Issue #404, opened 2017-10-18
Add startup diagnostics info


MarcinHoppe:
We can now capture [startup diagnostic information](https://docs.particular.net/nservicebus/hosting/?version=core_7#startup-diagnostics).

We need to decide what info is worth logging and implement it.

Connects to https://github.com/Particular/DotNetCoreLaunch/issues/78

## PoA

- [x] ~~Configured transactionality level (i.e. TransactionScope, SendsAtomicWithReceive, etc.)~~
   - Seems like this should be logged by Core https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs#L150
- [x] Native delayed-delivery (yes/no) + settings (batch size, interval, etc.)
   - [x] Timeout manager (enabled/disabled)
- [x] Circuit breaker settings
- [x] TTBR purge at startup (enabled/disabled)

ramonsmits:
Please execute `SELECT @@VERSION` on SQL Server to capture SQL Server version info. This helps us to understand if the customer is running with the latest patches on their SQL box.

weralabaj:
I propose to capture:
- Configured transactionality level (i.e. TransactionScope, SendsAtomicWithReceive, etc.)
- Native delayed-delivery (yes/no) + settings (batch size, interval, etc.)
- Timeout manager (enabled/disabled)
- Custom schema
- Multiple connection strings
- Circuit breaker settings
- TTBR purge at startup (enabled/disabled)

weralabaj:
@Particular/sqlserver-transport-maintainers Anything missing that should be included ☝️

SzymonPobiega:
Multiple connection strings are no longer supported but multiple schemas/catalogs should be reported. Question is how detailed should the report be i.e. if we should report yet/no or the actual configuration. The latter is accessible through https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Addressing/EndpointSchemaAndCatalogSettings.cs and https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Addressing/TableSchemaAndCatalogSettings.cs

weralabaj:
@SzymonPobiega @tmasternak We didn't document that diagnostic info output for sql persistence, I'm not sure if we should do that for sql server/both?
[...]

PR #448, opened 2018-06-13
Diagnostics info

weralabaj:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/404

### DIFF
--- a/src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
+++ b/src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
@@ -237,8 +237,21 @@ namespace NServiceBus.Transport.SQLServer
         {
             if (delayedDeliverySettings == null)
             {
+                settings.AddStartupDiagnosticsSection("NServiceBus.Transport.SqlServer.DelayedDelivery", new
+                {
+                    Native = false
+                });
                 return Task.FromResult(0);
             }
+
+            settings.AddStartupDiagnosticsSection("NServiceBus.Transport.SqlServer.DelayedDelivery", new
+            {
+                Native = true,
+                delayedDeliverySettings.Suffix,
+                delayedDeliverySettings.Interval,
+                BatchSize = delayedDeliverySettings.MatureBatchSize,
+                TimoutManager = delayedDeliverySettings.TimeoutManagerDisabled ? "disabled" : "enabled"
+            });
             var delayedMessageTable = CreateDelayedMessageTable();
             delayedMessageHandler = new DelayedMessageHandler(delayedMessageTable, CreateConnectionFactory(), delayedDeliverySettings.Interval, delayedDeliverySettings.MatureBatchSize);
             delayedMessageHandler.Start();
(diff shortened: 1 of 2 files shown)


src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
```diff
@@ -59,6 +63,10 @@ public static TransportExtensions<SqlServerTransport> UseSchemaForQueue(this Tra
             var schemasConfiguration = transportExtensions.GetSettings().GetOrCreate<QueueSchemaAndCatalogSettings>();
 
             schemasConfiguration.SpecifySchema(queueName, schema);
```
weralabaj: @SzymonPobiega @tmasternak I wonder if there's a better place to get that info from, ideal would be if we could have one place and yes/no section regarding using multiple conn strings, custom schemas, etc.
But I can't find a good place to put it in with access to settings (so I can add diagnostic info)... Any suggestions?

SzymonPobiega: Hmmm should we log all the queue/endpoint to schema mapping? I don't remember what we decided in SQLP? Do we log the custom schema or only the fact that there is a custom schema used?

weralabaj: only yes/no, because that may be considered as a first step towards leaking connection strings in logs 😂 
although we have time to revisit the idea

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -52,11 +56,24 @@ public override TransportReceiveInfrastructure ConfigureReceiveInfrastructure()
             {
                 scopeOptions = new SqlScopeOptions();
             }
+            else
+            {
+                settings.AddStartupDiagnosticsSection("NServiceBus.Transport.SqlServer.TransactionScope", new
+                {
+                    FeatureEnabled = true,
```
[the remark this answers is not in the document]
weralabaj: OK, I thought it's transaction scope specific, I'll revisit

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
SzymonPobiega: Then I would drop the whole thing entirely as it provides little value and adds noise if each endpoints has a dedicated schema.

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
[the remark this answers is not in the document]
weralabaj: seems like timeout and isolation level are always set (worst case, default values) so I'll report always

src/NServiceBus.SqlServer/SqlServerTransportSettingsExtensions.cs
weralabaj: removed

PR #450, opened 2018-06-18
Native transactions passed in SendOptions

### LINKED ISSUE #444: Allow passing an existing connection/transaction via send options
There are two scenarios that would greatly benefit from ability to pass `SqlConnection` and `SqlTransaction` through `SendOptions` to the transport

## Web store + send

In the web tier it is common to store some data to the database and send a message, atomically. In order to keep these two atomic, the data access operation and the send operation need to share the transaction context. Currently this is only possible by setting transaction mode to `TransactionScope`. This has drawbacks:
 * not supported in .NET Core 2.0
 * requires that the data access framework supports ambient transactions through `TransactionScope`
 * is not explicit -- less readable in the code

## High-throughput sends

If we want to just send 100s or 1000s of messages really fast the overhead of creating a `SqlConnection` for each send might be significant. In this scenario passing an already opened connection would help.

tmasternak:
Fixes #444

### DIFF
--- a/src/NServiceBus.SqlServer.AcceptanceTests/When_passing_custom_transaction_via_sendoptions.cs
+++ b/src/NServiceBus.SqlServer.AcceptanceTests/When_passing_custom_transaction_via_sendoptions.cs
@@ -1,0 +1,90 @@
+﻿namespace NServiceBus.SqlServer.AcceptanceTests
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Extensibility;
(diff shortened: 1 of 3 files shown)


SzymonPobiega:
@danielmarbach Looking at the code, if you use this in a handler it would substitute the existing transport transaction with the one you pass in so I guess this is reasonable. On the other hand, if used with immediate dispatch it would have no effect which is a bit surprising.

tmasternak:
I think we can detect if we are in a context of received message. In such case the transport will initialize `TransportTransaction` on receive. 

Issue created #452

danielmarbach:
You can also do a TryGet of the incoming message

PR #456, opened 2018-08-24
Update test projects to NServiceBus 7.1 and .NET Core 2.1

timbussmann:


timbussmann:
This solution needs definitely the patched versions as the solution references multiple NSB versions and will run into issues when trying to open the solution in VisualStudio.

SeanFeldman:
@Particular/sqlserver-transport-maintainers tests are failing on Full framework for Windows with ` Process exited with code -100`. I have a huntch it's related to TC project customizations. Could you please look into it or ping @Particular/deployment-maintainers? Thank you.

SzymonPobiega:
@SeanFeldman the .net core 2.1 output folder was not excluded from the test run, causing the .net core tests to be included which causes NUnit to go crazy. I've added the exclude and triggered the test build. Fingers crossed 🤞

timbussmann:
seems like that did the tricky @SzymonPobiega 👏

kbaley:
Can this be switched back to the template rather than using a custom build step? They look similar except that the template uses a wildcard to omit netcore folders which may have prevented the error.

SzymonPobiega:
@kbaley the other difference is that the non-template step uses `a` as a working directory which (after trying to swap) seems to be required by the compatibility tests (probably due to hard-coded paths). So can't switch right now.

PR #458, opened 2018-09-17
Fix `master` builds

tmasternak:
## Overview
There have been changes to `ApprovalTests` and `ApprovalUtilities` in versions 3.0.16. They are no longer targeting `.NETFramework 4.5` but `.NETFramework 4.6.1` and/or `.NETStandard 2.0`. That causes the build on master to fail for unit test which targets `.NETFramework 4.5`. 

In addition, this PR includes adding missing awaits in compatibility tests.

### DIFF
--- a/src/NServiceBus.SqlServer.UnitTests/NServiceBus.SqlServer.UnitTests.csproj
+++ b/src/NServiceBus.SqlServer.UnitTests/NServiceBus.SqlServer.UnitTests.csproj
@@ -13,8 +13,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="ApprovalTests" Version="3.*" NoWarn="NU1701" />
-    <PackageReference Include="ApprovalUtilities" Version="3.*" NoWarn="NU1701" />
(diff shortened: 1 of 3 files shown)


bording:
I know this probably won't change on master, but if you wanted the API approval test to also run on .NET Core, you'd need to lock to 3.0.13. 

However, the best thing to do would be to switch to Particular.Approvals anyway, which would happen on develop and not master.

src/NServiceBus.SqlServer.UnitTests/NServiceBus.SqlServer.UnitTests.csproj
```diff
@@ -13,8 +13,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="ApprovalTests" Version="3.*" NoWarn="NU1701" />
-    <PackageReference Include="ApprovalUtilities" Version="3.*" NoWarn="NU1701" />
+    <PackageReference Include="ApprovalTests" Version="3.0.15" NoWarn="NU1701" />
```
SzymonPobiega: So leaving it as is or changing to `3.0.13` as @bording suggested?

PR #466, opened 2018-10-18
Fix inspections

WojcikMike:
Change made to fix TC build


SzymonPobiega:
@WojcikMike I think the compilator error is because the TC build assigns it version 4.0 instead of 3.1.x. Feel free to merge.

PR #475, opened 2019-01-15
Use core 7.1 timeout migration mode

boblangley:
Switches the infrastructure timeout manager management to use the new core 7.1 migration mode


src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -23,13 +23,17 @@ internal SqlServerTransportInfrastructure(QueueAddressTranslator addressTranslat
             schemaAndCatalogSettings = settings.GetOrCreate<EndpointSchemaAndCatalogSettings>();
             delayedDeliverySettings = settings.GetOrDefault<DelayedDeliverySettings>();
             var timeoutManagerFeatureDisabled = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Disabled;
-            
+
+            var timeoutManagerDisabled = (delayedDeliverySettings != null && delayedDeliverySettings.TimeoutManagerDisabled) || timeoutManagerFeatureDisabled;
```
tmasternak: @boblangley can you give a bit more explanation on this logic? Especially the first part of the or 

> delayedDeliverySettings != null && delayedDeliverySettings.TimeoutManagerDisabled

what kind of scenario does this cover?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -23,13 +23,17 @@ internal SqlServerTransportInfrastructure(QueueAddressTranslator addressTranslat
...
+            settings.SetDefault(SettingsKeys.EnableMigrationMode, !timeoutManagerDisabled);
+
             diagnostics.Add("NServiceBus.Transport.SqlServer.TimeoutManager", new
             {
                 FeatureEnabled = !timeoutManagerFeatureDisabled
             });
-            
-            if (delayedDeliverySettings != null && timeoutManagerFeatureDisabled)
+
+            if (delayedDeliverySettings != null && timeoutManagerDisabled)
```
tmasternak: Similar to the above. What is the scenario in which `DelayedDeliverySettings` are not set?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
boblangley: Looking at it again it was too complicated. I reverted to closer to the original with the core migration mode call added as well as using the core settings extension to check feature status.

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
boblangley: See above

bording:
@boblangley I pushed up a couple of changes, so take a look and see what you think.

In general, I think there might be a startup check missing from the SQL Server native delay implementation that the others have: When using native delays and operating in migration mode, there is no startup check to verify that the Timeout Manager is actually active. Other transports had this check before moving to 7.1's migration mode feature, and have retained this after moving to it.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliverySettings.cs
```diff
@@ -39,7 +39,7 @@ public void BatchSize(int batchSize)
         /// </summary>
         public void DisableTimeoutManagerCompatibility()
```
SzymonPobiega: Should we mark this API as obsolete since now there is `DisableTimeoutManagerCompatibility` that people should be using?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -281,7 +286,7 @@ public override Task Start()
                 delayedDeliverySettings.Suffix,
                 delayedDeliverySettings.Interval,
                 BatchSize = delayedDeliverySettings.MatureBatchSize,
-                TimoutManager = delayedDeliverySettings.TimeoutManagerDisabled ? "disabled" : "enabled"
+                TimoutManager = delayedDeliverySettings.EnableMigrationMode ? "enabled" : "disabled"
```
SzymonPobiega: Should this be removed from SQLT diagnostics now and handled in the core?

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliverySettings.cs
bording: There is no other API, this is the API that needs to be used.

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
bording: This would still be showing whether the API has been called or not.

boblangley:
Changes look good Brandon

PR #480, opened 2019-03-13
Faster way to peek messages to retrieve message count

### LINKED ISSUE #481: SQL Server transport refuses to process messages if there are too many messages in the queue
SQL Server transport uses a periodic `select count(*)` query to asses how many messages is there in the queue to know how many receive tasks to spawn. If there is too many messages in the queue, `select count(*)` becomes a very slow operation requiring a table/index scan. It can easily take longer than 30 seconds in which case the operation times out and no messages are received.

The platform components are more likely to be affected than regular endpoints because they process messages coming from all endpoints e.g. the SC's audit processor or the Monitoring.

kbaley:
Fixes #481 

When peeking messages, we retrieve the count only up to the maximum concurrency level. This prevents the transport from effectively locking up if there are a lot of records to peek.

### DIFF
--- a/src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
+++ b/src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
@@ -17,7 +17,6 @@ namespace NServiceBus.Transport.SQLServer
 #pragma warning disable 618
             this.qualifiedTableName = qualifiedTableName;
             Name = queueName;
-            peekCommand = Format(SqlConstants.PeekText, this.qualifiedTableName, 100);
             receiveCommand = Format(SqlConstants.ReceiveText, this.qualifiedTableName);
             sendCommand = Format(SqlConstants.SendText, this.qualifiedTableName);
             purgeCommand = Format(SqlConstants.PurgeText, this.qualifiedTableName);
--- a/src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
+++ b/src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
@@ -17,7 +17,7 @@ namespace NServiceBus.Transport.SQLServer
 #pragma warning disable 618
             this.qualifiedTableName = qualifiedTableName;
             Name = queueName;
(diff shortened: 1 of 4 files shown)


src/NServiceBus.SqlServer/Queuing/TableBasedQueue.cs
```diff
@@ -17,7 +17,7 @@ public TableBasedQueue(string qualifiedTableName, string queueName)
 #pragma warning disable 618
             this.qualifiedTableName = qualifiedTableName;
             Name = queueName;
-            peekCommand = Format(SqlConstants.PeekText, this.qualifiedTableName);
+            peekCommand = Format(SqlConstants.PeekText, this.qualifiedTableName, 100);
```
SzymonPobiega: Should this be a function of the concurrency level? i.e. 10 times the max concurrency level so that there will be one peek query per 10 receives?

kbaley: This is a default value since at this time, we don't know what the max concurrency level is (do we?). The intent is to use the `FormatPeekCommand` to override it once you _do_ know the concurrency level.

SzymonPobiega: Ah, I didn't notice it. Then I'd say we should keep `peekCommand` not initialized until `FormatPeekCommand` is called to make sure that is is called in all cases it is needed.

src/NServiceBus.SqlServer/Receiving/MessagePump.cs
```diff
@@ -54,6 +54,7 @@ public async Task Init(Func<MessageContext, Task> onMessage, Func<ErrorContext,
 
         public void Start(PushRuntimeSettings limitations)
         {
+            inputQueue.FormatPeekCommand(limitations.MaxConcurrency);
```
SzymonPobiega: I think we should use a multiplier here for ensure that we don't call the count to frequently. My gut feeling says 10x multiplier would be fine. Plus we can cap it at say 100 so it would be `Math.Min(100, 10*maxConcurrency)`

PR #486, opened 2019-04-12
Update to new core recoverability seam

andreasohlund:


bording:
Shouldn't this be some sort of hotfix branch against `master` and not `develop`?

andreasohlund:
Since its only logging changes this didn’t warrant a hotfix

bording:
So it's not going to be released?

andreasohlund:
Correct, only updating develop to make sure the updated test passes

PR #491, opened 2019-06-11
Fixed startup check message and included original error.

SzymonPobiega:
Fixes https://github.com/Particular/KeepTheLightsOn/issues/269

I decided to reword the exception message and include the original error message instead of trying to be smart and parse the original exception message to detect the cause of the issue. 

The second options seemed to fragile as it would depend on the message of the exception thrown by the framework.

### DIFF
--- a/src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
+++ b/src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
@@ -189,9 +189,15 @@ namespace NServiceBus.Transport.SQLServer
                         scope.Complete();
                     }
                 }
-                catch (NotSupportedException)
+                catch (NotSupportedException ex)
                 {
-                    var message = "The version of System.Data.SqlClient in use does not support enlisting SQL connections in ambient transactions, so the TransactionScope transport transaction mode cannot be used. Use `EndpointConfiguration.UseTransport<SqlServerTransport>().Transactions` to select a different transport transaction mode.";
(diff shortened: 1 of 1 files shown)


src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -189,9 +189,15 @@ async Task<StartupCheckResult> CheckForAmbientTransactionEnlistmentSupport(SqlCo
                         scope.Complete();
                     }
                 }
-                catch (NotSupportedException)
+                catch (NotSupportedException ex)
                 {
-                    var message = "The version of System.Data.SqlClient in use does not support enlisting SQL connections in ambient transactions, so the TransactionScope transport transaction mode cannot be used. Use `EndpointConfiguration.UseTransport<SqlServerTransport>().Transactions` to select a different transport transaction mode.";
+                    var message = "The version of System.Data.SqlClient in use does not support one of the selected connection string options or " +
```
Scooletz: After thinking about this change, my opinion is that adding the original message is much more important than adding `or` and a new potential reason for this failure. LGTM

SzymonPobiega: So are you suggesting to revert the order here and start with the original message and then add our "commentary"?

Scooletz: Nope. Sorry for not being clear. What I meant is that adding was important and I don't care about the order. It's ready to be merged as it is IMO.

PR #494, opened 2019-06-14
Fix bad commit

andreasohlund:


bording:
Going to take a different approach to sort this out.

PR #495, opened 2019-06-14
Fixed startup check message and included original error.

bording:
The 4.1.3 release wasn't merged back into `develop`, so this PR bring the change to `master` now that we've changed to release flow.

CC: @SzymonPobiega


SzymonPobiega:
@andreasohlund @bording thanks! OK if I now merge it or do I have to do something else also?

andreasohlund:
Should be enough

Issue #463, opened 2018-09-24
Delayed messages can be delivered too early.


tmasternak:
## Problem

The t-sql command for inserting delayed messages is passed a due time which is calculated at the endpoint. However, when messages get moved to the input queue after maturing the value inserted is compared against `GETUTCDATE`. 
Any clock skews between SQL Server instance and the originating endpoint can cause messages to get delivered too early.

The command used for moving matured messages:

``` t-sql

DECLARE @NOCOUNT VARCHAR(3) = 'OFF';
IF ( (512 & @@OPTIONS) = 512 ) SET @NOCOUNT = 'ON';
SET NOCOUNT ON;

WITH message AS (
    SELECT TOP(@BatchSize) *
    FROM {0} WITH (UPDLOCK, READPAST, ROWLOCK)
    WHERE Due < GETUTCDATE())
DELETE FROM message
OUTPUT
    NEWID(),
    NULL,
    NULL,
    1,
    NULL,
    deleted.Headers,
    deleted.Body
INTO {1};

IF (@NOCOUNT = 'ON') SET NOCOUNT ON;
IF (@NOCOUNT = 'OFF') SET NOCOUNT OFF;

```

PR #496, opened 2019-07-04
Use TimeSpan instead of DateTime

WilliamBZA:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/463

By using a timespan, the date conversion is done on the SQL instance that will also be checking for Due messages. Because the clock calculating the due date is the same clock that will be checking the due date, this will prevent clock drift from occurring with delayed delivery.

### DIFF
--- a/src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
+++ b/src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
@@ -28,7 +28,7 @@ namespace NServiceBus.Transport.SQLServer
                     throw new Exception("Delayed delivery of messages with TimeToBeReceived set is not supported. Remove the TimeToBeReceived attribute to delay messages of this type.");
                 }
 
-                return (conn, trans) => delayedMessageTable.Store(operation.Message, behavior.DueTime, behavior.Destination, conn, trans);
+                return (conn, trans) => delayedMessageTable.Store(operation.Message, behavior.DueAfter, behavior.Destination, conn, trans);
             }
             return immediateDeliveryQueueOperationsReader.Get(operation);
         }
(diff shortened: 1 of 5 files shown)


src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageRow.cs
```diff
@@ -9,7 +9,7 @@ class DelayedMessageRow
     {
         DelayedMessageRow() { }
```
tmasternak: I propose to change `DelayedMessageRow` to something like `StoreDelayedMessageCommand` as this structure does not reflect anymore what is stored at the db level.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
```diff
@@ -37,11 +37,11 @@ static DispatchBehavior GetDueTime(UnicastTransportOperation operation)
         {
```
tmasternak: So just to make sure I'm no the same page: Right now the times for delayed delivery are `sender based`. Correct?

WilliamBZA: Uhm, not really. If a delay is specified with a `DelayDeliveryWith ` then the delayed delivery is based on the SQL instance. If the delay is specified with a `DoNotDeliverBefore`, then it is sender clock based. But even in that scenario, the code calculates the duration difference at the sender point between the sender's current time and when the message should be delivered. This duration is passed to the Sql instance, and the difference is used to calculate the delivery date there. There could maybe be µs differences depending on how long it takes for the sender to send the insert command to the Sql database, but that's about it.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageRow.cs
WilliamBZA: :+1: Renamed

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeiveryTableBasedQueueFactory.cs
tmasternak: @WilliamBZA ok but both for `DelayDeliveryWith` and `DoNotDeliverBefore` the only piece of data we send to the SQL Instance is `dueAfter`, right?

WilliamBZA: @tmasternak yes, which is a duration. Not a specific point in time, which means if the sender has clock drift of 2 hours ahead of the Sql server and it requests a timeout to occur in the next 15 minutes, the timeout will still be in 15 minutes. The clock drift won't matter.

PR #497, opened 2019-07-08
Warn about timeout manager

WilliamBZA:
Warns for https://github.com/Particular/NServiceBus.SqlServer/issues/468 if `TimeoutManager` is used.

### DIFF
--- a/src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
+++ b/src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
@@ -57,6 +57,21 @@
             Assert.IsEmpty(context.Logs.Where(l => l.Message.Contains("Current configuration of the endpoint uses the TimeoutManager feature for delayed delivery - an option which is not recommended for new deployments. SqlTransport native delayed delivery should be used instead. It can be enabled by calling `UseNativeDelayedDelivery()`.")));
         }
 
+        [Test]
+        public async Task Should_not_warn_when_both_native_delayed_delivery_and_timeoutmanage_is_configured_with_compatibility_disabled()
+        {
+            Requires.MessageDrivenPubSub();
+
+            var context = await Scenario.Define<ScenarioContext>()
+                .WithEndpoint<EndpointWithTimeoutManagerAndNativeEnabledButCompatibilityDisabled>()
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            Assert.True(context.EndpointsStarted, "because it should not prevent endpoint startup");
+
+            Assert.IsEmpty(context.Logs.Where(l => l.Message.Contains("Current configuration of the endpoint uses the TimeoutManager feature for delayed delivery - an option which is not recommended for new deployments. SqlTransport native delayed delivery should be used instead. It can be enabled by calling `UseNativeDelayedDelivery()`.")));
+        }
+
         public class EndpointWithTimeoutManagerAndNotNative : EndpointConfigurationBuilder
         {
             public EndpointWithTimeoutManagerAndNotNative()
@@ -71,8 +86,8 @@
             {
                 EndpointSetup<DefaultServer>(config =>
                 {
-                    config.UseTransport<SqlServerTransport>().UseNativeDelayedDelivery();
                     config.EnableFeature<TimeoutManager>();
+                    config.UseTransport<SqlServerTransport>().UseNativeDelayedDelivery();
                 });
             }
         }
@@ -83,6 +98,20 @@
             {
                 EndpointSetup<DefaultServer>(config =>
                 {
+                    var settings = config.UseTransport<SqlServerTransport>().UseNativeDelayedDelivery();
+                    settings.DisableTimeoutManagerCompatibility();
+                });
+            }
+        }
+
+        public class EndpointWithTimeoutManagerAndNativeEnabledButCompatibilityDisabled : EndpointConfigurationBuilder
+        {
+            public EndpointWithTimeoutManagerAndNativeEnabledButCompatibilityDisabled()
+            {
+                EndpointSetup<DefaultServer>(config =>
+                {
+                    config.EnableFeature<TimeoutManager>();
+
                     var settings = config.UseTransport<SqlServerTransport>().UseNativeDelayedDelivery();
                     settings.DisableTimeoutManagerCompatibility();
                 });
--- a/src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
+++ b/src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
@@ -1,0 +1,91 @@
+﻿namespace NServiceBus.AcceptanceTests.NativeTimeouts
+{
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using Logging;
+    using NUnit.Framework;
+    using Transport.SQLServer;
+
+    public class When_configuring_delayed_delivery : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_warn_when_timeoutmanager_is_configured_without_native_delayed_delivery()
+        {
+            Requires.MessageDrivenPubSub();
+
+            var context = await Scenario.Define<ScenarioContext>()
+                .WithEndpoint<EndpointWithTimeoutManagerAndNotNative>()
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            Assert.True(context.EndpointsStarted, "because it should not prevent endpoint startup");
+
+            var log = context.Logs.Single(l => l.Message.Contains("Current configuration of the endpoint uses the TimeoutManager feature for delayed delivery - an option which is not recommended for new deployments. SqlTransport native delayed delivery should be used instead. It can be enabled by calling `UseNativeDelayedDelivery()`."));
+            Assert.AreEqual(LogLevel.Warn, log.Level);
+        }
+
+        [Test]
+        public async Task Should_not_warn_when_timeoutmanager_and_native_delayed_delivery_are_both_configured()
+        {
+            Requires.MessageDrivenPubSub();
+
+            var context = await Scenario.Define<ScenarioContext>()
+                .WithEndpoint<EndpointWithTimeoutManagerAndNative>()
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            Assert.True(context.EndpointsStarted, "because it should not prevent endpoint startup");
+
+            Assert.IsEmpty(context.Logs.Where(l => l.Message.Contains("Current configuration of the endpoint uses the TimeoutManager feature for delayed delivery - an option which is not recommended for new deployments. SqlTransport native delayed delivery should be used instead. It can be enabled by calling `UseNativeDelayedDelivery()`.")));
+        }
+
+        [Test]
+        public async Task Should_not_warn_when_only_native_delayed_delivery_is_configured()
+        {
+            Requires.MessageDrivenPubSub();
+
+            var context = await Scenario.Define<ScenarioContext>()
+                .WithEndpoint<EndpointWithOnlyNative>()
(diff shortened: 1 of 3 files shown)


src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -24,6 +25,11 @@ internal SqlServerTransportInfrastructure(QueueAddressTranslator addressTranslat
             delayedDeliverySettings = settings.GetOrDefault<DelayedDeliverySettings>();
             var timeoutManagerFeatureDisabled = !settings.IsFeatureEnabled(typeof(TimeoutManager));
 
+            if (!timeoutManagerFeatureDisabled)
+            {
+                Logger.Warn("TimeoutManager feature has been enabled. SqlTransport supports delayed delivery natively and it is recommended to use the Native Timeout Manager instead.");
```
tmasternak: Should we mention `UseNativeDelayedDelivery();` in the warn statement?

``` suggestion
Logger.Warn("Current configuration of the endpoint uses TimeoutManager feature for delayed delivery - an option which is not recommended for new deployments. SqlTransport native delayed delivery should be used instead. It can be enabled by calling `UseNativeDelayedDelivery` method.");
```

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -24,6 +25,11 @@ internal SqlServerTransportInfrastructure(QueueAddressTranslator addressTranslat
             delayedDeliverySettings = settings.GetOrDefault<DelayedDeliverySettings>();
             var timeoutManagerFeatureDisabled = !settings.IsFeatureEnabled(typeof(TimeoutManager));
 
+            if (!timeoutManagerFeatureDisabled)
```
bording: Not sure this is the right place to be checking this. It's too early to check the final state of the feature. Instead, this should be done as a startup check. See https://github.com/Particular/NServiceBus.RabbitMQ/blob/master/src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs#L65 and https://github.com/Particular/NServiceBus.RabbitMQ/blob/master/src/NServiceBus.Transport.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs#L81 for examples of what I'm talking about.

WilliamBZA: I've moved it to the startup task. I was debating whether or not it belonged here or later, so thanks for chiming in. I joined this taskforce exactly to learn about things like this.

I wonder if the existing check should also be moved later then though?

The part where the infrastructure adds to the diagnostics. This is only used in debugging purposes, but if it could be changed after this point, should the diagnostics be set later too?

src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs
WilliamBZA: Changed!

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryInfrastructure.cs
```diff
@@ -1,17 +1,27 @@
...
 {
+    using Features;
+    using Logging;
     using Settings;
 
     static class DelayedDeliveryInfrastructure
     {
         public static StartupCheckResult CheckForInvalidSettings(SettingsHolder settings)
         {
+            var timeoutManagerEnabled = settings.IsFeatureEnabled(typeof(TimeoutManager));
```
bording: Checking for `Enabled` isn't the right thing to be doing here. You should be checking to see if the feature is active instead:

https://github.com/Particular/NServiceBus.RabbitMQ/blob/master/src/NServiceBus.Transport.RabbitMQ/DelayedDelivery/DelayInfrastructure.cs#L84

WilliamBZA: I've changed this check to use the FeatureState

WilliamBZA:
Check always runs now, and the logic in the check only takes the send only codepath if DelayedDelivery is enabled.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryInfrastructure.cs
```diff
@@ -1,17 +1,32 @@
...
     using Settings;
 
     static class DelayedDeliveryInfrastructure
     {
         public static StartupCheckResult CheckForInvalidSettings(SettingsHolder settings)
         {
-            var sendOnlyEndpoint = settings.GetOrDefault<bool>("Endpoint.SendOnly");
-            if (sendOnlyEndpoint)
+            var timeoutManagerEnabled = settings.GetOrDefault<FeatureState>(typeof(TimeoutManager).FullName) == FeatureState.Active;
+            if (timeoutManagerEnabled)
```
bording: Taking a step back and looking at what the overall goal of this change is, I'm not sure this the right way to check if the warning should be displayed.

We want to point people to use native delayed delivery instead of the timeout manager, correct? Even when they call `UseNativeDelayedDelivery`, the timeout manager will still be active and running in migration mode. They have to also call `DisableTimeoutManagerCompatibility` to have the timeout manager completely turned off.

However, I don't think this warning is intended be displayed when they are running in migration mode, correct? That's an expected reason for the timeout manager to be active.

In that case, we don't care about the state of the timeout manager at all. All we need to do is display the warning when `UseNativeDelayedDelivery` hasn't been called, meaning when `delayedDeliverySettings is null`.

WilliamBZA: I've added a few tests to check the warning state here. It now only logs when `TimeoutManager` is active, but without native delivery.

src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
```diff
@@ -0,0 +1,91 @@
...
+            }
+        }
+
+        public class EndpointWithOnlyNative : EndpointConfigurationBuilder
+        {
+            public EndpointWithOnlyNative()
+            {
+                EndpointSetup<DefaultServer>(config =>
+                {
+                    config.UseTransport<SqlServerTransport>().UseNativeDelayedDelivery();
```
bording: We expect customers to use the `DisableTimeoutManagerCompatibility` API to get themselves into this state, so I suggest adding a call to that and enabling the timeout manager to make this test closer to the real scenario.

Or perhaps add an additional test for the combo I'm describing.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryInfrastructure.cs
```diff
@@ -1,17 +1,34 @@
...
+                var sendOnlyEndpoint = settings.GetOrDefault<bool>("Endpoint.SendOnly");
+                if (sendOnlyEndpoint)
+                {
+                    return StartupCheckResult.Failed("Native delayed delivery is only supported for endpoints capable of receiving messages.");
+                }
             }
+            else
+            {
+                var timeoutManagerEnabled = settings.IsFeatureActive(typeof(TimeoutManager));
+                if (timeoutManagerEnabled)
```
bording: I was originally thinking checking if the feature was active wasn't necessary, but users could disable the timeout manager   and not have native delays, making their endpoint not have any delay capability, so it does make sense to keep it!

I suppose you could add a test for this scenario, but since the timeout manager is disabled by default in the acceptance tests, all the existing tests cover this already.

src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
WilliamBZA: I've added the call to disabled the timeout manager compatibility. I don't know if we need a dedicated test for this state though.

Currently the code already disables compatibility (https://github.com/Particular/NServiceBus.SqlServer/pull/497/files#diff-734321b72d52bc31733b86191d75e96cL36), which means the codepath being tested is the same either way. I'm happy to still add the test though if you'd prefer? Being explicit about it is probably not a bad thing.

src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
```diff
@@ -83,7 +83,8 @@ public EndpointWithOnlyNative()
             {
                 EndpointSetup<DefaultServer>(config =>
                 {
-                    config.UseTransport<SqlServerTransport>().UseNativeDelayedDelivery();
+                    var settings = config.UseTransport<SqlServerTransport>().UseNativeDelayedDelivery();
```
bording: This should still have `config.EnableFeature<TimeoutManager>();` as well. Otherwise it's not actually a useful test.

src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
bording: The code you've linked to is a fallback mechanism to detect if the feature has been disabled for some reason, and if so, don't require the API to be explicitly called as well.

I still think it makes sense to test the actual expected scenario of the feature being active and the API having been called.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryInfrastructure.cs
tmasternak: @bording I think that this check was also added to enable warning the users that their configuration is not the recommended one i.e. TimeoutManager and not Native Delayed Delivery.

src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
tmasternak: Why isn't this useful in current form?

bording: Because it's not actually testing the scenario it's claiming to test. You should be testing that the `UseNativeDelayedDelivery` API actually results in the Timeout manager not being active, but the acceptance tests disable the timeout manager by default, so this test would always pass. You need to enable the feature and call `UseNativeDelayedDelivery` to test the actual behavior of an endpoint when the API is called.

src/NServiceBus.SqlServer/DelayedDelivery/DelayedDeliveryInfrastructure.cs
bording: @tmasternak I think you might be misunderstanding my comment.

src/NServiceBus.SqlServer.AcceptanceTests/NativeTimeouts/When_configuring_delayed_delivery.cs
WilliamBZA: The tests here are explicitly about testing the warn messages produced with a given setup. This specific test (`Should_not_warn_when_only_native_delayed_delivery_is_configured`) is testing that there is no warning when native delivery is configured and the timeout manager is disabled explicitly (by calling `.DisableFeature` - which the ATT framework does). By calling `config.EnableFeature<TimeoutManager>();` the test will not be testing that scenario. It would in fact be testing the warning messages in a migration scenario where existing timeouts are being drained (which is tested by the `Should_not_warn_when_timeoutmanager_and_native_delayed_delivery_are_both_configured` test)

The Active state of features is also controlled by Core, and not by the downstream transport. This means that if we add a test to check the `Active` state of a feature, we are in fact testing [core logic](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs#L22) and not transport logic.

WilliamBZA:
@bording ok to merge?

bording:
It appears that TC is refusing to build the new changes, saying they are outdated.

WilliamBZA:
Closed and replaced with #498

Issue #504, opened 2019-09-19
Not possible to delay a message more than int.Max milliseconds


Rensvind:
https://github.com/Particular/NServiceBus.SqlServer/blob/cf1d3f87ad81518bd6c4850df7b5dcfed6c00360/src/NServiceBus.SqlServer/Queuing/SqlConstants.cs#L54

It is currently not possible to delay a message further in time than int.Max milliseconds since you are using DATEADD(ms, @DueAfterMs, GETUTCDATE()));

This works
`sendOptions.DelayDeliveryWith(TimeSpan.FromMilliseconds(int.MaxValue));`

This 
`sendOptions.DelayDeliveryWith(TimeSpan.FromMilliseconds((long)int.MaxValue + 1));`

fails with 

> System.Data.SqlClient.SqlException: 'Arithmetic overflow error converting expression to data type int.
> The statement has been terminated.'
>

ramonsmits:
> The number argument cannot exceed the range of int.

https://docs.microsoft.com/en-us/sql/t-sql/functions/dateadd-transact-sql?redirectedfrom=MSDN&view=sql-server-2017

That is a limitation on the DATEADD function.

andreasohlund:
@Rensvind we've confirmed that this is a bug and also made it a priority fix.

Since this was introduce in v4.3.0 would you be able to go back to v4.1.3 for now?

Rensvind:
Great!

Yeah, it's no problem rolling back to an earlier version.

danielmarbach:
Will be contained in 4.3.1 that is going to be released soonish

danielmarbach:
We pushed out the release https://discuss.particular.net/t/nservicebus-sqlserver-4-3-1-patch-release-available/1437

PR #507, opened 2019-10-16
Fix DATEADD overflow by passing timespan parts to SQL layer

danielmarbach:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/504
Alternative to https://github.com/Particular/NServiceBus.SqlServer/pull/508

The previous maintenance release switched from `DateTime` to `TimeSpan`. This PR breaks the `TimeSpan` into its corresponding parts and passes those parts to the underlying SQL layer that then uses those to do `DATEADD` calculations. 

The benefit of this approach is the following:

- The TimeSpan is preserved
- The TimeSpan parts are ints
- There is no interpretation in the SQL layer needed thus no logic is done in the SQL and the parameters can be passed as is
- It is aligned with how the [DATEADD function](https://docs.microsoft.com/en-us/sql/t-sql/functions/dateadd-transact-sql) works because it operates on date parts
- It is aligned with the current design because the transport seems to do the logic in the translation layers in code where needed and not necessarily in SQL
- According to profiling no impact on the performance

### DIFF
--- a/src/NServiceBus.SqlServer/DelayedDelivery/StoreDelayedMessageCommand.cs
+++ b/src/NServiceBus.SqlServer/DelayedDelivery/StoreDelayedMessageCommand.cs
@@ -27,7 +27,11 @@
         {
             AddParameter(command, "Headers", SqlDbType.NVarChar, headers);
             AddParameter(command, "Body", SqlDbType.VarBinary, bodyBytes);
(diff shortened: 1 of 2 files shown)


tmasternak:
Looked at this and #508. I can't say that I have a strong preference - this is helpful isn't  it :wink:. That said I see an advantage in this one in that all the logic is in the same place in C#. In #508 it's in C# and it t-sql. 

So if I really need to choose one I would go with this one.

PR #502, opened 2019-09-11
Add delay in MoveMaturedDelayedMessages on exception

christoffer-rydberg:
Fixes #501

PR #509, opened 2019-10-16
Always delay moving of delayed messages even when exceptions occur to honor the ProcessingInterval

### LINKED ISSUE #501: No delay when moving matured delayed messages after exception
I'm working in a project where our logs sometimes get spammed with "Exception thrown while performing cancellation". It happens in [DelayedMessageHandler](https://github.com/Particular/NServiceBus.SqlServer/blob/54fadb504f3b0752c748010e51504cb62bf54b38/src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageHandler.cs#L63) when the sql transaction commit throws an exception. This results in tens of thousands of log messages. The issue seems to be that the while loop continues without delay.

danielmarbach:
Fixes #501 
Alternative to https://github.com/Particular/NServiceBus.SqlServer/pull/502

This alternative makes sure that on the cancellation case the `Task.Delay` throwing an `OperationCanceledException` would not make the outer task `Cancelled` and thus bomb on shutting down

### DIFF
--- a/src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageHandler.cs
+++ b/src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageHandler.cs
@@ -14,6 +14,7 @@ namespace NServiceBus.Transport.SQLServer
             this.connectionFactory = connectionFactory;
             this.interval = interval;
             this.batchSize = batchSize;
+            message = $"Scheduling next attempt to move matured delayed messages to input queue in {interval}";
         }
 
         public void Start()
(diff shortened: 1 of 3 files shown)


src/NServiceBus.SqlServer/DelayedDelivery/DelayedMessageHandler.cs
```diff
@@ -47,8 +47,6 @@ async Task MoveMaturedDelayedMessages()
                             transaction.Commit();
                         }
                     }
-                    Logger.DebugFormat("Scheduling next attempt to move matured delayed messages to input queue in {0}", interval);
-                    await Task.Delay(interval, cancellationToken).ConfigureAwait(false);
                 }
                 catch (OperationCanceledException)
```
danielmarbach: This could actually go now because the semantics of the code has changed. I figured I'll leave it in case we ever introduce command cancellation

PR #512, opened 2019-10-16
 Always delay moving of delayed messages even when exceptions occur to honor the ProcessingInterval

danielmarbach:
Backport of #509


danielmarbach:
The linux build never worked, see https://builds.particular.net/viewType.html?buildTypeId=NServiceBusSqlServer_33&tab=buildTypeHistoryList&state=failed&branch_NServiceBusSqlServer=__all_branches__ This needs to be force merged

PR #513, opened 2019-10-16
Create and reset scripts

danielmarbach:


danielmarbach:
@andreasohlund  I added you as a reviewer to show to you that this PR now turns the 3.x branch green again on the build server

andreasohlund:
So this was the thing that broke the Linux build?

danielmarbach:
If by this you mean the not existing Create scripts then yes

PR #518, opened 2019-10-25
Native Pub-Sub

mikeminutillo:
Doco https://github.com/Particular/docs.particular.net/pull/4583


mikeminutillo:
Switching to non-draft more to get the build server picking it up

src/NServiceBus.SqlServer.UnitTests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -78,13 +109,14 @@ EXEC sp_releaseapplock @Resource = '{0}_lock'";
...
         public static NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> UseCatalogForEndpoint(this NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> transportExtensions, string endpointName, string catalog) { }
         public static NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> UseCatalogForQueue(this NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> transportExtensions, string queueName, string catalog) { }
         public static NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> UseCustomSqlConnectionFactory(this NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> transportExtensions, System.Func<System.Threading.Tasks.Task<System.Data.SqlClient.SqlConnection>> sqlConnectionFactory) { }
-        public static NServiceBus.Transport.SQLServer.DelayedDeliverySettings UseNativeDelayedDelivery(this NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> transportExtensions) { }
```
mikeminutillo: Should this be obsoleted?

src/NServiceBus.SqlServer/PubSub/SubscriptionTable.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            using (new TransactionScope(TransactionScopeOption.Suppress, TransactionScopeAsyncFlowOption.Enabled))
+            {
+                using (var connection = await connectionFactory.OpenNewConnection().ConfigureAwait(false))
+                using (var command = connection.CreateCommand())
+                {
+                    command.CommandText = getSubscribersCommand;
+                    for (var i = 0; i < topics.Length; i++)
+                    {
+                        command.Parameters.Add($"Topic_{i}", SqlDbType.VarChar).Value = topics[i];
```
mikeminutillo: :+1:

This won't work if you have a single message going to >2098 topics. That seems unlikely in real-world use-cases.

src/NServiceBus.SqlServer/PubSub/PolymorphicSubscriptionStore.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        IEnumerable<string> GetTopicsFor(Type messageType)
+        {
+            return GetMessageHierarchy(messageType).Select(TopicName.From);
+        }
+
+        IEnumerable<Type> GetMessageHierarchy(Type messageType)
+        {
+            return messageHierarchyCache.GetOrAdd(messageType, t => GenerateMessageHierarchy(t).ToArray());
```
mikeminutillo: The Topic naming strategy and the type hierarchy don't change at runtime. We could cache the topics instead of the types so that we don't have to do the conversion on every publish.

Although we probably don't need to as the results are cached for 5 seconds by default anyway. But a customer still has the option to disable caching.

src/NServiceBus.SqlServer/PubSub/PolymorphicSubscriptionStore.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        static IEnumerable<Type> GenerateMessageHierarchy(Type messageType)
+        {
+            var t = messageType;
+            while (t != null)
```
mikeminutillo: This means we'll always try to publish to subscribers to `System.Object` even though there is likely never going to be a subscriber for that topic. Is that an issue? Should that be allowed?

src/NServiceBus.SqlServer/PubSub/CachedSubscriptionStore.cs
```diff
@@ -0,0 +1,68 @@
...
+                _ => new CacheItem
+                {
+                    Stored = DateTime.UtcNow,
+                    Subscribers = inner.GetSubscribers(eventType)
+                });
+
+            var age = DateTime.UtcNow - cacheItem.Stored;
+            if (age >= cacheFor)
+            {
+                cacheItem.Subscribers = inner.GetSubscribers(eventType);
```
mikeminutillo: It's possible that two (or more) threads are doing this at the same time (i.e. Both Get, Both expire, Both retrieve from the database). Should the above call become an `AddOrUpdate` and include this logic in the update?

src/NServiceBus.SqlServer.UnitTests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -64,29 +84,37 @@ CREATE NONCLUSTERED INDEX [Index_Due] ON {0}
...
         public static NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> UseSchemaForEndpoint(this NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> transportExtensions, string endpointName, string schema) { }
         public static NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> UseSchemaForQueue(this NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> transportExtensions, string queueName, string schema) { }
         public static NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> WithPeekDelay(this NServiceBus.TransportExtensions<NServiceBus.SqlServerTransport> transportExtensions, System.Nullable<System.TimeSpan> delay = null) { }
     }
+    public class SubscriptionSettings
+    {
+        public SubscriptionSettings() { }
```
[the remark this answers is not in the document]
SzymonPobiega: Technically not but I don't think this could be a problem. In the core some of the setting classes have internal constructors (Outbox) and some have public (Routing). I would keep it public to keep the code simpler.

src/NServiceBus.SqlServer/PubSub/CachedSubscriptionStore.cs
SzymonPobiega: I think it can stay this way. Worst case both threads would try to update the properties of the same `cacheItem` object but this should not be a big deal since there is no danger of producing non-consistent results.

src/NServiceBus.SqlServer/PubSub/PolymorphicSubscriptionStore.cs
SzymonPobiega: My guess is for the scenario when cache is disabled the the cost of going to the DB to fetch the subscriptions is orders of magnitude larger than the reflection here. But the the engineer in me does not allow me to pass by such a comment ;-) I'll add the cache ❤️

src/NServiceBus.SqlServer/PubSub/SubscriptionSettings.cs
```diff
@@ -0,0 +1,62 @@
...
+        TimeSpan? TimeToCacheSubscriptions;
+
+        /// <summary>
+        /// Overrides the default name for the subscription table. All endpoints in a given system need to agree on that name in order for them to be able
+        /// to subscribe to and publish events.
+        /// </summary>
+        /// <param name="tableName">Name of the table.</param>
+        /// <param name="schemaName">Schema in which the table is defined if different from default schema configured for the transport.</param>
+        /// <param name="catalogName">Catalog in which the table is defined if different from default catalog configured for the transport.</param>
+        public void SubscriptionTableName(string tableName, string schemaName = null, string catalogName = null)
```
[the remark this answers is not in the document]
SzymonPobiega: We'll end up with two separate tables. The dilemma here is that if we tried to prevent this (e.g. by requiring explicit configuration or by defaulting to dbo for the subscription table) we would hurt people who use same (but non-default) schema in all endpoints e.g. `nsb`. I believe this is much more popular approach than using schema-per-endpoint so I think we should keep it that way (and risk people using not getting pub sub to work).

src/NServiceBus.SqlServer/PubSub/TopicNames.cs
```diff
@@ -0,0 +1,9 @@
+namespace NServiceBus.Transport.SQLServer
+{
+    using System;
+
+    static class TopicName
+    {
+        public static string From(Type type) => $"{type.Namespace}.{type.Name}";
```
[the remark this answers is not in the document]
SzymonPobiega: Hmmm I actually wonder why didn't we use `Type.FullName`? I think I just copied it from RabbitMQ transport code. The downside of this approach is that nested types look like regular types here so we can't distinguish between two nested types if they have the same name but are nested in different classes. What do you think? Should we change it to `FullName`?

src/NServiceBus.SqlServer/Queuing/SqlConstants.cs
```diff
@@ -255,5 +255,71 @@ SELECT t.name
 WHERE c.object_id = OBJECT_ID('{0}')
     AND c.name = 'Headers'";
 
+        public static readonly string CreateSubscriptionTableText = @"
```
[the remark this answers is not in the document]
SzymonPobiega: I think this is a result of a mess. I'll align it all to use `public static readonly string` as the most common one.

src/NServiceBus.SqlServer/Receiving/QueueCreator.cs
```diff
@@ -31,6 +33,13 @@ public async Task CreateQueueIfNecessary(QueueBindings queueBindings, string ide
                 }
                 transaction.Commit();
             }
+
+            using (var connection = await connectionFactory.OpenNewConnection().ConfigureAwait(false))
+            using (var transaction = connection.BeginTransaction())
```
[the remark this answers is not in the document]
SzymonPobiega: Result of me copy-pasting stuff between classes. Yes, we can use the same one. I'll fix it.

src/NServiceBus.SqlServer/Sending/OperationSorter.cs
```diff
@@ -0,0 +1,93 @@
...
+        public SortingResult(IEnumerable<UnicastTransportOperation> defaultDispatch, IEnumerable<UnicastTransportOperation> isolatedDispatch)
+        {
+            DefaultDispatch = defaultDispatch;
+            IsolatedDispatch = isolatedDispatch;
+        }
+    }
+
+    static class OperationSorter
+    {
+        public static SortingResult SortAndDeduplicate(this IEnumerable<UnicastTransportOperation> source, QueueAddressTranslator addressTranslator)
```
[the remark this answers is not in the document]
SzymonPobiega: If we define the comparer of `UnicastTransportOperation` then it would have to invoke `addressTranslator.Parse` for each comparison :( We would have to pre-calculate that so we are stuck with another type anyway.

src/NServiceBus.SqlServer/SubscriptionTableInstaller.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// in the container via a convention so the only way to pass objects to them is via settings or via container. The transport infrastructure
+    /// cannot access the container so we're doing it via settings (HACK).
+    /// </summary>
+    class SubscriptionTableInstaller : INeedToInstallSomething
+    {
+        SubscriptionTableCreator creator;
+
+        public SubscriptionTableInstaller(ReadOnlySettings settings)
+        {
+            creator = settings.GetOrDefault<SubscriptionTableCreator>();
```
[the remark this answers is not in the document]
SzymonPobiega: Good catch. This would likely throw exception if a SQL Server transport dll is referenced but an endpoint uses a different transport ❤️

SzymonPobiega:
@mikeminutillo I pushed the fixes

src/NServiceBus.SqlServer/PubSub/PolymorphicSubscriptionStore.cs
SzymonPobiega: I did some research and conducted interrogations and it seems like native pub sub transports like RabbitMQ or Learning support subscribing to `IEvent` but not `object`. Message-driven pub sub supports neither. I think we should support `IEvent`. I'll update the code.

SzymonPobiega:
Closing in favour of #527 .

PR #524, opened 2019-10-29
Setting LangVersion to 7.3

HEskandari:


HEskandari:
@danielmarbach these are as close as it can get to a green build. There seems to be a brittle Windows test that breaks the build.

HEskandari:
@danielmarbach this one is green too.

PR #541, opened 2020-01-23
[SPIKE] Introduce legacy project with if def

danielmarbach:
Alternative to #540


danielmarbach:
Compatibility tests are failing because they haven't been maintained for a while. But the PR demonstrates that we can execute have compatibility of the client as well as execute all the tests against both flavors of the transport without a problem

danielmarbach:
Outcome will go into https://github.com/Particular/NServiceBus.SqlServer/pull/543

PR #580, opened 2020-02-25
Guard against queue creation race

danielmarbach:


danielmarbach:
Had to do the same in NHibernate https://github.com/Particular/NServiceBus.NHibernate/blob/49c34a319d612d53ea5a1c41b7235c50f1849322/src/NServiceBus.NHibernate/TimeoutPersisters/OptimizedSchemaUpdate.cs#L107

PR #581, opened 2020-02-28
Use correct ifdef for project's target frameworks

bording:
Looks like we overlooked this one when updating things for Microsoft.Data.SqlClient support.


bording:
Once this is merged, we should bring it over to `release-6.0` as well.

PR #602, opened 2020-04-24
Cannot pass Timeout.InfiniteTimeSpan

ramonsmits:
Disable the native delayed delivery processing by passing an infinite timeout value.


mauroservienti:
@ramonsmits is this still relevant?

mauroservienti:
@ramonsmits can you also please describe a use case for this?

WilliamBZA:
Replaced by https://github.com/Particular/NServiceBus.SqlServer/pull/721

ramonsmits:
Reopening as this is needed when there are multiple instances.

WilliamBZA:
> Customers can also achieve something similar by changing the timeout check interval.

PR #612, opened 2020-05-19
Faster way to peek messages to retrieve message count (#480)

### LINKED ISSUE #481: SQL Server transport refuses to process messages if there are too many messages in the queue

lailabougria:
Fixes #481

When peeking messages, we retrieve the count only up to the maximum concurrency level. This prevents the transport from effectively locking up if there are a lot of records to peek.

### DIFF
--- a/src/NServiceBus.SqlServer.IntegrationTests/When_message_receive_takes_long.cs
+++ b/src/NServiceBus.SqlServer.IntegrationTests/When_message_receive_takes_long.cs
@@ -53,6 +53,7 @@
         static async Task TryPeekQueueSize(TableBasedQueue tableBasedQueue)
         {
             await ExecuteInTransactionScope(async c => {
+                tableBasedQueue.FormatPeekCommand(100);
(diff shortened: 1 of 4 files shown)


lailabougria:
@WojcikMike found out that the test failing in this PR was also failing in the base branch (release-3.1) so we don't think the changes made here had any negative impact in this case.

WojcikMike:
Yeah, I saw that too. @bording Please have a look at the branch. @lailabougria can you change it from draft to a regular one?

bording:
Based on the conversation around the test, I'll go ahead and merge this.

PR #617, opened 2020-05-26
Added obsolete to indicate that there is a newer package

ramonsmits:


bording:
I don't think we should do this. This is not considered obsolete. We are intentionally shipping both packages.

Until we need to take a dependency on a feature that is only offered by `Microsoft.Data.SqlClient` (which could happen eventually), there's no reason to stop shipping both packages.

PR #632, opened 2020-06-09
Update to NServiceBus Core 7.3.0

andreasohlund:


andreasohlund:
@bording FYI this removed the compat tests as we have discussed in the past

Issue #452, opened 2018-06-27
Prevent `UseCustomSqlConnectionAndTransaction` usage in the message receive context.


tmasternak:
Usage of `UseCustomSqlConnectionAndTransaction` should be allowed only outside of the Pipeline (and as a result of any handler, behavior etc.).

Currently users might get unexpected behavior as a result of combining custom transactions with e.g. `ImmediateDispatch` or configured receive TransactionMode.

PR #634, opened 2020-06-16
CustomTransaction should always immediately dispatch

dvdstelt:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/452

The custom TransportTransaction is overwritten before the message is dispatched if it is set from inside the pipeline.

Instead we decided that if a message is sent immediately (immediate dispatch) it will be dispatched, but not yet committed into the database until the custom transaction by the user is committed. Or rolled back, obviously.


tmasternak:
I think we need to change the [Dispatch](https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.Transport.SqlServer/Sending/MessageDispatcher.cs#L38) method logic as well.

Currently, when we dispatch in the isolated mode we always create a new transactional context discarding the one passed by the user.

src/NServiceBus.Transport.SqlServer/PublishOptionsExtensions.cs
```diff
@@ -20,8 +20,13 @@ public static class PublishOptionsExtensions
         /// <param name="transaction">SqlTransaction instance that will be used by any operations performed by the transport.</param>
         public static void UseCustomSqlTransaction(this PublishOptions options, SqlTransaction transaction)
         {
+            // When dispatching, the TransportTransaction is overwritten.
+            // The only way for a custom transaction to work is by using immediate dispatch and messages should only appear when the user commits the custom transaction.
+            // Which is exactly what will happen after NServiceBus dispatches this message immediately.
+            options.RequireImmediateDispatch();
+
             var transportTransaction = new TransportTransaction();
-            transportTransaction.Set(SettingsKeys.TransportTransactionSqlConnectionKey, transaction.Connection);
```
SzymonPobiega, line 24: Although it is not technically part of public API, users may have been depending on the fact that the connection is part of the TransportTransaction. It is true that it is not strictly necessary but I'd leave it just in case, for compat reasons.

dvdstelt:
@SzymonPobiega I added it back

PR #680, opened 2020-09-11
Endpoint column cannot be null

mikeminutillo:
Rasied https://github.com/Particular/docs.particular.net/pull/5001


DavidBoike (approved):

SzymonPobiega (approved):

DavidBoike:
I would merge it now but should it get included with the next release's notes?

SimonCropp:
where is the upgrade guide on how people should fix this for existing tables?

DavidBoike:
IIRC there's no upgrade guidance necessary. This is how SQL was interpreting it anyway due to the primary key (see original issue https://github.com/Particular/docs.particular.net/issues/4993) so it's just an adjustment to reality.

PR #708, opened 2020-11-03
Validate database connectivity at endpoint startup

WilliamBZA:
Two commits so we can separate the connection check from the behaviour change if we want.


tmasternak:
Added my comments to the `master` PR.

tmasternak (changes requested):
Same as for the PR against `master`

WilliamBZA:
Messed that Up real bad. Closing.

PR #714, opened 2020-11-05
Change read code to support Always Encrypted

WilliamBZA:
I am not sure of the implications of this and would need someone more well versed in the history behind the need for sequential reads.

The reasons this is required are:

* When reading an encrypted column, `CommandBehavior.SequentialAccess` is not supported
* Without `CommandBehavior.SequentialAccess`, `SqlDataReader.GetStream` no longer works

This means that if we want to support having the headers and body columns and always encrypted columns, we need to stop using `CommandBehavior.SequentialAccess`.

Stopping the use of `CommandBehavior.SequentialAccess` means we have to stop using `GetStream`. Does typecasting to `byte[]` always work? I can't find references for when it would break down.

The PR only changes the code path when the `Column Encryption Setting` connection string parameter is enabled

### DIFF
--- a/src/NServiceBus.Transport.SqlServer/Queuing/MessageRow.cs
+++ b/src/NServiceBus.Transport.SqlServer/Queuing/MessageRow.cs
@@ -48,7 +48,6 @@
 
         static async Task<MessageRow> ReadRow(SqlDataReader dataReader)
         {
-            //HINT: we are assuming that dataReader is sequential. Order or reads is important !
             return new MessageRow
             {
                 id = await dataReader.GetFieldValueAsync<Guid>(0).ConfigureAwait(false),
@@ -95,15 +94,9 @@
             }
         }
 
-        static async Task<byte[]> GetBody(SqlDataReader dataReader, int bodyIndex)
+        static Task<byte[]> GetBody(SqlDataReader dataReader, int bodyIndex)
         {
-            // Null values will be returned as an empty (zero bytes) Stream.
-            using (var outStream = new MemoryStream())
-            using (var stream = dataReader.GetStream(bodyIndex))
-            {
-                await stream.CopyToAsync(outStream).ConfigureAwait(false);
-                return outStream.ToArray();
-            }
+            return Task.FromResult((byte[])dataReader[bodyIndex]);
         }
 
         static async Task<T> GetNullableAsync<T>(SqlDataReader dataReader, int index) where T : class
--- a/src/NServiceBus.Transport.SqlServer/Queuing/MessageRow.cs
+++ b/src/NServiceBus.Transport.SqlServer/Queuing/MessageRow.cs
@@ -96,6 +96,7 @@
 
         static async Task<byte[]> GetBody(SqlDataReader dataReader, int bodyIndex)
         {
+            // Null values will be returned as an empty (zero bytes) Stream.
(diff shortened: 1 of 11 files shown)


src/NServiceBus.Transport.SqlServer/Queuing/MessageRow.cs
```diff
@@ -95,15 +94,9 @@ static async Task<string> GetHeaders(SqlDataReader dataReader, int headersIndex)
             }
         }
 
-        static async Task<byte[]> GetBody(SqlDataReader dataReader, int bodyIndex)
+        static Task<byte[]> GetBody(SqlDataReader dataReader, int bodyIndex)
         {
-            // Null values will be returned as an empty (zero bytes) Stream.
-            using (var outStream = new MemoryStream())
```
WilliamBZA, line 101: The original comment `// We need sequential access to not buffer everything into memory` wasn't true anyway, as the entire byte array is read into memory in this method regardless of the sequential access flag - otherwise how would it return the byte array?

src/NServiceBus.Transport.SqlServer/Queuing/TableBasedQueue.cs
```diff
@@ -54,7 +55,7 @@ public virtual async Task<MessageReadResult> TryReceive(SqlConnection connection
         {
             using (var command = new SqlCommand(receiveCommand, connection, transaction))
             {
-                return await ReadMessage(command).ConfigureAwait(false);
+                return await ReadMessage(command, isStreamSupported).ConfigureAwait(false);
```
tmasternak, line 58: Do we need to pass `isStreamSupported`? Can we use the instance field?

src/NServiceBus.Transport.SqlServer/Queuing/TableBasedQueue.cs
```diff
@@ -70,17 +71,22 @@ public Task Send(OutgoingMessage message, TimeSpan timeToBeReceived, SqlConnecti
             return SendRawMessage(messageRow, connection, transaction);
         }
 
-        static async Task<MessageReadResult> ReadMessage(SqlCommand command)
+        static async Task<MessageReadResult> ReadMessage(SqlCommand command, bool isStreamSupported)
```
tmasternak, line 74: Can we make this instance method and remove the `isStreamSupported` argument?

WilliamBZA: This is a static method and won't have access to instance methods

src/NServiceBus.Transport.SqlServer/Queuing/TableBasedQueue.cs
WilliamBZA: Not unless we change `ReadMessage` to be an instance method instead of static

tmasternak: I would lean towards making it instance. Currently `command` and `isStreamSupported` have different lifecycles. `Command` changes per-call and the other one stays the same for the whole endpoint operation.

WilliamBZA: Done

mauroservienti (approved):

PR #721, opened 2020-11-17
Expose disable delayed delivery setting

WilliamBZA:
Exposes a setting to disable delayed delivery infrastructure.


tmasternak (approved):

mauroservienti (approved):

WilliamBZA:
Withdrawing. This is a very dangerous change as Saga timeouts and delayed retries depend on timeouts too. By not creating the infrastructure we could end up with customers getting strange exceptions because of the relationship between the various pieces.

Customers can also achieve something similar by changing the timeout check interval.

ramonsmits:
> Saga timeouts and delayed retries depend on timeouts too.

THey are dependant on delayed delivery. If a transport doesn't support this an exception MUST be raised if core request a timeout (sagas) or requests delayed delivery with the options. WHich it does, as when an endpoint has delayed delivery disabled it should not report back this capability to core.

PR #725, opened 2020-11-26
Better SchemaVerification index search

carstengehling:
SchemaVerification currently just searches for indexes on the queue table named `Index_Expires` and `Index_RowVersion` (the latter with a specific type).

Our deployment scripts create the correct indexes, but they are named differently. This gives false warnings on startup.

In this PR I have created queries that instead looks for an index, that have the specified column as its first element (key_ordinal=1), and for RowVersion still also checkes for non-clustered type. Apart from solving our own problem, it also solves the issue, that startup currently will not warn, if eg. Index_Expires exists but is designed wrongly.

PR #728, opened 2020-12-02
Better SchemaVerification index search

WilliamBZA:
Cherry-pick of #725 for 6.2

SchemaVerification currently just searches for indexes on the queue table named Index_Expires and Index_RowVersion (the latter with a specific type).

Our deployment scripts create the correct indexes, but they are named differently. This gives false warnings on startup.

In this PR I have created queries that instead looks for an index, that have the specified column as its first element (key_ordinal=1), and for RowVersion still also checkes for non-clustered type. Apart from solving our own problem, it also solves the issue, that startup currently will not warn, if eg. Index_Expires exists but is designed wrongly.


WilliamBZA:
Backport to release branch, I'm not sure if we need this...

SzymonPobiega:
Not sure either, but in case we do, it looks good ;)

SzymonPobiega (approved):

PR #732, opened 2020-12-15
Fix API Approvals

SeanFeldman:


bording:
> Core expects a task and does not handle a null response here.

??

SeanFeldman:
That's a GitHub issue that adds a title/comment when changing a branch unrelated to what the branch is doing... fixed the description.

bording (approved):

PR #741, opened 2021-01-19
change connection scope when using scoped trans

SimonCropp:
@SzymonPobiega regarding this https://discuss.particular.net/t/this-platform-does-not-support-distributed-transactions/1807/3

```
using (var conn = OpenConnection("DB2"))
using (var conn = OpenConnection("DB3"))
{
   //Do work in DB 2
   //Do work in DB 3
}

even if both connection string are identical you will get a DTC escalation.
```

would this PR reduce the likelihood of DTC escalation?

### DIFF
--- a/src/NServiceBus.Transport.SqlServer/Receiving/ProcessWithTransactionScope.cs
+++ b/src/NServiceBus.Transport.SqlServer/Receiving/ProcessWithTransactionScope.cs
@@ -25,15 +25,15 @@
                     using (var connection = await connectionFactory.OpenNewConnection().ConfigureAwait(false))
                     {
                         message = await TryReceive(connection, null, receiveCancellationTokenSource).ConfigureAwait(false);
+                    }
 
-                        if (message == null)
-                        {
-                            // The message was received but is not fit for processing (e.g. was DLQd).
-                            // In such a case we still need to commit the transport tx to remove message
-                            // from the queue table.
-                            scope.Complete();
-                            return;
-                        }
+                    if (message == null)
+                    {
+                        // The message was received but is not fit for processing (e.g. was DLQd).
+                        // In such a case we still need to commit the transport tx to remove message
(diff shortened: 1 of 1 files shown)


SimonCropp:
note that the diff is easier to grok in split view

SzymonPobiega:
I am pretty sure that `Close` is equivalent to `Dispose` when it comes to this behavior. To my knowledge the magic that allows you to open two connections sequentially within a scope without causing the escalation is done by using a dedicated connection pool area/markers. When a connection that is enlisted in a scope is disposed (or closed, I believe), it is not returned to the pool but rather "left aside" and when you try to created/open a new one in the same scope, you end up using the same underlying connection so for the SQL Server it it is still the same connection/transaction context.

Have you verified that the `Close` before `TryProcess` is not enough to return the connection to that special state and `Dispose` is needed? By the way, I am just asking because I like your PR from the style and readability point of view so we would like to merge it anyway but I'd like to understand if we should try to release it soon?

SimonCropp:
> I am pretty sure that Close is equivalent to Dispose when it comes to this behavior

i wasnt sure if that was the case. so decided to be certain about it

> Have you verified that the Close before TryProcess is not enough to return 

No. i am having trouble nailing down the issue. we have the sql transport, sql persistence, sql attachments and EF in the mix. so i am trying to rule out some potential issues. i think this is another one https://github.com/Particular/NServiceBus.Persistence.Sql/issues/594

SzymonPobiega:
@SimonCropp given that we identified the root cause of the problem and have issues open for fixing it, would it be OK if I closed this one? We are going to adopt this approach in the next major version of the transport but since it does not change the actual behvaior I don't see a reason to change the version that targets NServiceBus 7.

SimonCropp:
@SzymonPobiega yep. thanks for all the help

PR #790, opened 2021-03-11
Add Core v7 configuration API

tmasternak:


src/NServiceBus.Transport.SqlServer/SqlServerTransport.cs
```diff
@@ -46,6 +46,15 @@ public SqlServerTransport(Func<CancellationToken, Task<SqlConnection>> connectio
             ConnectionFactory = connectionFactory;
         }
 
+        /// <summary>
+        /// Used for backwards compatibility with the legacy transport api.
+        /// </summary>
+        internal SqlServerTransport()
```
SzymonPobiega, line 52: How about moving it to `obsoletes` (need to make this class partial)

src/NServiceBus.Transport.SqlServer/SqlServerTransport.cs
```diff
@@ -76,6 +87,21 @@ public override async Task<TransportInfrastructure> Initialize(HostSettings host
             return infrastructure;
         }
 
+        void ValidateConfiguration()
```
SzymonPobiega, line 90: How about moving it to `obsoletes` once this class is partial.

SzymonPobiega (approved):

mauroservienti (approved):

PR #773, opened 2021-02-26
SQL Transport - Support CancellationTokens, core alpha 891

DavidBoike:

PR #793, opened 2021-03-15
avoid object disposed exception in MessageReciver.Stop

adamralph:
relates to https://github.com/Particular/NServiceBus.SqlServer/pull/773

Since we dispose the token on line 108.

### DIFF
--- a/src/NServiceBus.Transport.SqlServer/Receiving/MessageReceiver.cs
+++ b/src/NServiceBus.Transport.SqlServer/Receiving/MessageReceiver.cs
@@ -78,18 +78,20 @@
         {
             messageReceivingCancellationTokenSource?.Cancel();
 
-            cancellationToken.Register(() => messageProcessingCancellationTokenSource?.Cancel());
-
-            await messageReceivingTask.ConfigureAwait(false);
-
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Transport.SqlServer/Receiving/MessageReceiver.cs
```diff
@@ -78,18 +78,20 @@ public async Task StopReceive(CancellationToken cancellationToken = default)
         {
             messageReceivingCancellationTokenSource?.Cancel();
 
-            cancellationToken.Register(() => messageProcessingCancellationTokenSource?.Cancel());
+            using (cancellationToken.Register(() => messageProcessingCancellationTokenSource?.Cancel()))
```
andreasohlund, line 81: Ah, nice, we should do this in MSMQ (and I guess the other transports as well) right?

andreasohlund (approved):

src/NServiceBus.Transport.SqlServer/Receiving/MessageReceiver.cs
adamralph: I'm going through and doing all of them.

Issue #837, opened 2021-05-25
Update Microsoft.Data.SqlClient package to v2


rickdsw:
In the current latest version of the NServiceBus.Transport.SqlServer package, Microsoft.Data.SqlClient v1.1.3 is used. This version of Microsoft.Data.SqlClient package contains an issue for Dotnet Framework, because of the way the SNI.dll is installed. In order to make that package work, you need to install the package in every project which contains a soft reference to the package. 

More details about this can be found in this issue:
https://github.com/dotnet/SqlClient/issues/361#issuecomment-569026464

A fix for this issue is added to version v2, where they change the path of the SNI.dll.
Force installing the Microsoft.Data.SqlClient package v2 in my project, does not solve the issue, since NServiceBus.Transport.SqlServer still references v1.1.3. This results in the following exception:
Unable to load DLL 'Microsoft.Data.SqlClient.SNI.x86.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)

I was hoping you could create a new version for the NServiceBus.Transport.SqlServer package which uses Microsoft.Data.SqlClient >=v2.0.0

ramonsmits:
Automatic deployment of transitive dependencies requires SDK style projects. Old style VS projects require the "host" project to explicitly select the transitive dependencies to ensure that during compile these will be added to the output folder.

By default VS will use SDK style projects only for .net core projects but it is possible to target .NET Framework using SDK style projects. All our projects use SDK style projects too, for example the project file of this package: https://github.com/Particular/NServiceBus.SqlServer/blob/b7cdb1e3cad153c8fa064b0bfaba920a89c26647/src/NServiceBus.SqlServer/NServiceBus.SqlServer.csproj

See [Target frameworks in SDK-style projects](https://docs.microsoft.com/en-us/dotnet/standard/frameworks)

We will not update the dependency to >=v2.0.0 to ensure compatibility with v1.x as there are no breaking API changes.

PR #838, opened 2021-05-25
Update Microsoft.Data.SqlClient to 2.1.3

ramonsmits:
Resolves https://github.com/Particular/NServiceBus.SqlServer/issues/837


ramonsmits:
As explained in #837 this is not a bug that requires fixing and is by design.

PR #846, opened 2021-06-08
Defend against TryPeek command returning null

adamralph:


src/NServiceBus.Transport.SqlServer/Queuing/TableBasedQueue.cs
```diff
@@ -37,7 +37,7 @@ public virtual async Task<int> TryPeek(SqlConnection connection, SqlTransaction
                 CommandTimeout = timeoutInSeconds ?? 30
             })
             {
-                var numberOfMessages = (int)await command.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false);
+                var numberOfMessages = (int)(await command.ExecuteScalarAsync(cancellationToken).ConfigureAwait(false) ?? 0);
```
tmasternak, line 40: When can this return `null`?

adamralph: During connectivity problems, apparently.

adamralph:
https://github.com/Particular/NServiceBus.SqlServer/pull/847 is probably better

src/NServiceBus.Transport.SqlServer/Queuing/TableBasedQueue.cs
tmasternak: 👍

Can we add a comment in the conde or add that in the PR description so that we don't loose this context?

adamralph: I'm not sure it's really required. `ExecuteScalarAsync` returns an `object` which can obviously be `null`. This code should never have assumed that the return value is castable to an `int`. But see https://github.com/Particular/NServiceBus.SqlServer/pull/847/files which I think should replace this PR.

DavidBoike:
Task force for the enhancement release decided to take this simpler version. The TryPeek only cares about are there some messages or not. If a connectivity issue says there we can't figure there are any messages, that's fine. Also would be weird for a method named TryX to throw something.

DavidBoike (approved):

Issue #848, opened 2021-06-08
ExecuteScalarAsync calls are at risk of null refs or casting exceptions due to database connectivity problems and/or over-assumption


adamralph:
For example, https://github.com/Particular/NServiceBus.SqlServer/blob/eac7963bcd1130d7abfeb24aac8d12be5f5bc15c/src/NServiceBus.Transport.SqlServer/Queuing/TableBasedQueue.cs#L40 can throw a `NullReferenceException` because the value returned by `ExecuteScalarAsync` can be `null` either during connectivity problems or if a mistake has been made elsewhere in the code (e.g. a bad SQL statement). These calls should not assume that the return value can be cast to an `int`.

Possible solutions:

- https://github.com/Particular/NServiceBus.SqlServer/pull/846
- https://github.com/Particular/NServiceBus.SqlServer/pull/847

PR #847, opened 2021-06-08
defend against TryPeek command returning unexpected value

adamralph:
alternative to https://github.com/Particular/NServiceBus.SqlServer/pull/846

PR #851, opened 2021-06-09
defend against ExecuteScalarAsync returning unexpected values

adamralph:
@kbaley @DavidBoike @kentdr this is a more generalised version of https://github.com/Particular/NServiceBus.SqlServer/pull/846 which also takes into account the recommendation to log warnings by @tmasternak and also covers the remaining calls to `ExecuteScalarAsync` as mentioned by @ramonsmits [here](https://github.com/Particular/NServiceBus.SqlServer/pull/847/files#r647179404) and [here](https://github.com/Particular/NServiceBus.SqlServer/pull/847/files#r647180005).

fixes https://github.com/Particular/NServiceBus.SqlServer/issues/848


ramonsmits (approved):
Not a fan of even having the null check on this edge case but no harm in checking for null for clarity.

Issue #751, opened 2021-02-05
SQL Server transport does not allow to easily override the catalog for all queues/endpoints


SzymonPobiega:
SQL Server transport uses the catalog from the connection string (`initial catalog` property) when accessing DB objects. This can be overridden on per-endpoint or per-queue basis using `UseCatalogPerQueue` nad `UseCatalogPerEndpoint`. These APIs allow users to set up a _catalog-per-endpoint_ topology in which each endpoint uses a separate catalog for data (business + queues).

It does not allow, however, to set up a _queues-in-shared-db_ topology easily because all destination queues have to be configured via these APIs. Event worse, if a new endpoint is added, all other endpoints need to be re-configured and re-deployed.

A solution to this problem might be an API similar to `DefaultSchema` e.g. `DefaultCatalog`. When provided, the transport would use that catalog instead of the one from the connection string. This allows the business code to reach its DB objects in the business DB while SQLT uses objects from `DefaultCatalog`.

SzymonPobiega:
@SimonCropp does it describe the problem/solution well enough?

tmasternak:
Another approach to solving the problem would be to:
 * Use endpoint in `TransactionScope` mode
 * Use endpoint specific connection string per endpoint e.g. "Initial Catalog=Orders"
 * Register custom connection factories for SqlT and SqlP with a callback that uses the default connection string (to enable pooling) but change the database after opening. This will work as the connection pooling invokes `sp_reset_connection` when the connection gets back to the pool and this brings back the initial catalog to the one specified in the connection string. E.g.
 
``` csharp
transport.UseCustomSqlConnectionFactory(
    sqlConnectionFactory: async () =>
    {
        var connection = new SqlConnection("SomeConnectionString");
        await connection.OpenAsync().ConfigureAwait(false);
        connection.ChangeDatabaseAsync("NserviceBus");
    });
```

SimonCropp:
@tmasternak 

> Use endpoint in TransactionScope mode

did u smoke test this approach?

SimonCropp:
@SzymonPobiega 

> does it describe the problem/solution well enough?

yes i think so

tmasternak:
@SimonCropp what I've tested is that the connection which gets returned to the pool has its initial catalog reset before it's reused.

SimonCropp:
[...]

PR #853, opened 2021-06-10
Added DefaultCatalog property to override the connection string

kentdr:
This is to address [issue 751](https://github.com/Particular/NServiceBus.SqlServer/issues/751)


src/NServiceBus.Transport.SqlServer.UnitTests/SqlServerTransportTests.cs
```diff
@@ -36,5 +38,36 @@ public void It_accepts_connection_string_with_catalog_property(string connection
...
+        public void It_overrides_catalog_with_default_catalog(string connectionString, string defaultCatalog)
+        {
+            var transport = new SqlServerTransport(connectionString)
+            {
+                DefaultCatalog = defaultCatalog
+            };
+
+            transport.ParseConnectionAttributes();
+
+            Assert.AreEqual(defaultCatalog, transport.Catalog);
```
kbaley, line 54: Is it worth asserting they _aren't_ equal beforehand to be safe?

kentdr: From my testing, `transport.Catalog` is null until `ParseConnectionAttributes` is called.  `transport.Catalog` is never set to the invalid value the whole time, it just goes from null to the Default.  Do you think we should assert that `transport.Catalog` is null before we `ParseConnectionAttributes`?

kbaley: To be safe, yeah. Otherwise there's no guarantee they aren't equal _before_ the call to `ParseConnectionAttributes`

src/NServiceBus.Transport.SqlServer.AcceptanceTests/MultiCatalog/SqlUtilities.cs
```diff
@@ -0,0 +1,72 @@
...
+    using Microsoft.Data.SqlClient;
+#endif
+    using System;
+    using System.Threading.Tasks;
+    using System.Threading;
+
+    public static class SqlUtilities
+    {
+        public static async Task<bool> CheckIfTableExists(string catalogName, string schemaName, string tableName, SqlConnection connection, CancellationToken cancellationToken = default)
+            => await RunCommand(connection, async (command, token) =>
```
DavidBoike, line 15: Avoid an async/await state machine by directly returning the task
```suggestion
        public static Task<bool> CheckIfTableExists(string catalogName, string schemaName, string tableName, SqlConnection connection, CancellationToken cancellationToken = default)
            => RunCommand(connection, async (command, token) =>
```

Same in DropTable below.

DavidBoike (approved):

kbaley (approved):

PR #860, opened 2021-06-17
Better DTC check, doesn't even open connection on netcore (6.3)

DavidBoike:


kentdr (approved):

src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -281,13 +287,13 @@ async Task<StartupCheckResult> TryEscalateToDistributedTransactions(TransactionO
                                   "Note that different transaction modes may affect consistency guarantees as you can't rely on distributed " +
                                   "transactions to atomically update the database and consume a message. Original error message: " + ex.Message;
                 }
-                catch (SqlException sqlException)
+                catch (Exception exception)
                 {
                     message = "Could not escalate to a distributed transaction while configured to use TransactionScope. Check original error message for details. " +
                                   "In case the problem is related to distributed transactions you can still use SQL Server transport but " +
                                   "should specify a different transaction mode via `EndpointConfiguration.UseTransport<SqlServerTransport>().Transactions`. " +
```
kentdr, line 294:
```suggestion
                                  "should specify a different transaction mode by setting the `TransportTransactionMode` property on the `SqlServerTransport` instance when configuring the endpoint. " +
```
WIth the 8.0 API changes `UseTransport()` returns the routing settings so the transaction mode has to be set directly on the transport instance.  The above sentence a bit less specific than it was prior and possibly confusing.  Maybe we should [send them to the docs](https://docs.particular.net/transports/transactions#transactions-transaction-levels-supported-by-nservicebus-transports) instead?

DavidBoike: We don't link in docs from components because we have no good way to ensure those links stay valid. But this is the 6.3 version so the UseTransport<T> form is correct.

DavidBoike: But I am going to steal your wording (it's a bit hand-wavy but it's accurate) for the OTHER pull request on master where the message was not updated by the transport seam TF.

kbaley (approved):

PR #829, opened 2021-05-06
GitHub Actions - master

DavidBoike:

PR #886, opened 2021-09-16
Create CI workflow

timbussmann:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          fetch-depth: 0
+      - name: Install SQL Server (Windows)
+        if: matrix.name == 'Windows'
+        run: |
+          choco install sql-server-express
+          # Set necessary SQLCMD environment variable for the setup step
+          echo "SQLCMDSERVER=.\SQLEXPRESS" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
+          # Set connection string as environment variable for the tests
+          echo "SqlServerTransportConnectionString=Data Source=.\SQLEXPRESS;Initial Catalog=nservicebus;Integrated Security=True" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
```
timbussmann, line 37: while it seems to be default value, it might be reasonable to explicitly define that we want to run with the developer version of the SQL Server (for licensing reasons) by using the `MSSQL_PID` configuration environment variable. See https://hub.docker.com/_/microsoft-mssql-server

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Start SQL Server (Linux)
+        if: matrix.name == 'Linux'
+        run: |
+          sa_pw=DEVELOPERS!developers!
+          docker run -d -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=$sa_pw" -p 1433:1433 --name sqlserver mcr.microsoft.com/mssql/server:2017-latest
```
timbussmann, line 42: while it seems to be default value, it might be reasonable to explicitly define that we want to run with the developer version of the SQL Server (for licensing reasons) by using the MSSQL_PID configuration environment variable. See https://hub.docker.com/_/microsoft-mssql-server

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          choco install sql-server-express
+          # Set necessary SQLCMD environment variable for the setup step
+          echo "SQLCMDSERVER=.\SQLEXPRESS" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
+          # Set connection string as environment variable for the tests
+          echo "SqlServerTransportConnectionString=Data Source=.\SQLEXPRESS;Initial Catalog=nservicebus;Integrated Security=True" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
+      - name: Start SQL Server (Linux)
+        if: matrix.name == 'Linux'
+        run: |
+          sa_pw=DEVELOPERS!developers!
```
DavidBoike, line 41: Should make this a secret, just in this repository is fine.

With a different, randomized value.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Build
+        run: dotnet build src --configuration Release
+      - name: Upload packages
+        if: matrix.name == 'Windows'
+        uses: actions/upload-artifact@v2.2.2
+        with:
+          name: NuGet packages
+          path: nugets/
+          retention-days: 7
+      - name: Prepare SQL Server
```
DavidBoike, line 66: For completeness, in my attempt SqlCmd wasn't going to work (can't remember why, related to ACI maybe) and I had built this tool https://github.com/Particular/Particular.GitHubActionsTool - if it's not needed, great! But it also delegated out to a SQL script that I think is already in the repo, which might be helpful since I think SqlCmd can do that too.

Here's where I called the tool:
https://github.com/Particular/NServiceBus.SqlServer/pull/829/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR88

.github/workflows/ci.yml
soujay: Replaced it with a guid generator

timbussmann: what's the point of making it a secret?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+            name: Windows
+          - os: ubuntu-20.04
+            name: Linux
+      fail-fast: false
+    steps:
+      - name: Check for secrets
+        env:
+          SECRETS_AVAILABLE: ${{ secrets.SECRETS_AVAILABLE }}
+        shell: pwsh
+        run: exit $(If ($env:SECRETS_AVAILABLE -eq 'true') { 0 } Else { 1 })
```
timbussmann, line 33: can we remove this secrets check again if we're not using secrets?

DavidBoike: Yes, and the `pull_request_target` trigger would need to come out too.

Issue #947, opened 2021-12-08
I seems, that a message should be put into a dead letter queue if QueueNotFoundException occurs


MikhailIvanou:
Hi,
We have a bit unusual case that is connected with our database restore process between different test/dev environments, anyway our **NSB endpoint ends up processing the same message in a loop**.

As far as i understand, the problem is that **"NServiceBus.SqlServer.ForwardDestination"** header of that message **points to a queue that has been removed** and doesn't exist anymore.

I would like to see something like that:

```
    abstract class ReceiveStrategy
    {
       
           ...

        protected async Task<Message> TryReceive(SqlConnection connection, SqlTransaction transaction, CancellationTokenSource receiveCancellationTokenSource)
        {
            var receiveResult = await InputQueue.TryReceive(connection, transaction).ConfigureAwait(false);

            if (receiveResult.IsPoison)
            {
                await ErrorQueue.DeadLetter(receiveResult.PoisonMessage, connection, transaction).ConfigureAwait(false);
                return null;
            }

            if (receiveResult.Successful)
            {
                try
                {
                  if (await TryHandleDelayedMessage(receiveResult.Message, connection, transaction).ConfigureAwait(false))
                  {
                      return null;
                  }
                }
                catch(QueueNotFoundException ex)
                {
                  ...
                  await ErrorQueue.DeadLetter(receiveResult.Message, connection, transaction).ConfigureAwait(false);
                  ...
                }

                return receiveResult.Message;
            }
            receiveCancellationTokenSource.Cancel();
            return null;
        }
```

What do you think about it?

SzymonPobiega:
Hi

Yes, indeed the forwarding mechanism should include the code like this one. I'll add a label to the issue to mark it as a candidate for the next release. Thanks for reporting it! In the meantime do you need help in unblocking the endpoint? It should be fine to just create the destination queue manually and let the message go there to fix the immediate problem. Then you can move the message to any queue you like via a T-SQL.

MikhailIvanou:
Nice to hear, thanks for response. Yes, we've already fixed it temporarily.

SzymonPobiega:
@MikhailIvanou this has been fixed and will be released with version 7.0

PR #952, opened 2021-12-29
Handle delayed message with invalid destination

SzymonPobiega:
by moving them to the error queue with all the exception headers set and the FailedQ header pointing to the (invalid) destination queue


SzymonPobiega:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/947

tmasternak (approved):

Issue #954, opened 2022-01-14
Messages may be lost in cases of database connectivity problems


MarcWils:
### Symptoms

Messages are removed from the queue but message handlers are not executed for them and these messages are not audited -- they are lost leaving no trace.

### Who's affected

Affected are possibly all users of SQL Server transport but the issue has only been confirmed in one environment that uses Azure SQL.

### Original description

We're in the middle of investigating a performance issue. The Azure SQL database containing all of our queue tables hits a 100% CPU spike for a long period of time. This results is many timeouts, transactions errors and so. That is to be expected and something we're trying to figure out. However, it seems that some messages were "lost" during that time. There's no sign these messages were put aside for delayed delivery or moved to the error queue.

We use a recent version of NServiceBus v7 and NServiceBus.SqlServer v6.3 (System.Data.SqlClient). Endpoints are running in transaction mode TransactionScope and use native delayed delivery. Exceptions bubble up to the NServiceBus infrastructure triggering immediate and possibly delayed retries.

Message loss is something we've never seen before and only seem to occasionally happen during the Azure SQL CPU spike. As I said, there are many timeout exceptions, transaction in doubt exceptions and so on. But I can't relate them to the messages we've lost.

When digging through the logs and code, I've noticed this issue #848. Our logs contains multiple warning "peekCommand returned a null value." during the time of database problems. I assume SQL actually returns a result for the peek query, but the result is lost. I was wondering if the same issue might occur when receiving a message from the input queue. A SqlDataReader is used to read the message: https://github.com/Particular/NServiceBus.SqlServer/blob/ca40c86287618aefa76bce8c45df93aa5cd6bea9/src/NServiceBus.Transport.SqlServer/Queuing/TableBasedQueue.cs#L83-L88

That's basically the same construction the ExecuteScalar method uses: https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/System.Data/System/Data/SqlClient/SqlCommand.cs#L1162-L1171

What if there's an actual scenario where the .ReadAsync() method returns `false`, but SQL did delete and return a message from the queue. The transaction will committed and no warning/error is logged: https://github.com/Particular/NServiceBus.SqlServer/blob/ca40c86287618aefa76bce8c45df93aa5cd6bea9/src/NServiceBus.Transport.SqlServer/Receiving/ProcessWithTransactionScope.cs#L28-L35

[...]

PR #961, opened 2022-02-09
Transport transaction not committed when no message received

tmasternak:
Connected to: #954

### DIFF
--- a/src/NServiceBus.Transport.SqlServer/Receiving/ProcessStrategy.cs
+++ b/src/NServiceBus.Transport.SqlServer/Receiving/ProcessStrategy.cs
@@ -40,28 +40,21 @@
 
         public abstract Task ProcessMessage(CancellationTokenSource stopBatchCancellationTokenSource, CancellationToken cancellationToken = default);
 
-        protected async Task<Message> TryGetMessage(SqlConnection connection, SqlTransaction transaction, CancellationTokenSource stopBatchCancellationTokenSource, CancellationToken cancellationToken = default)
+        protected async Task<MessageReadResult> TryGetMessage(SqlConnection connection, SqlTransaction transaction, CancellationTokenSource stopBatchCancellationTokenSource, CancellationToken cancellationToken = default)
         {
             var receiveResult = await inputQueue.TryReceive(connection, transaction, cancellationToken).ConfigureAwait(false);
 
             if (receiveResult.IsPoison)
             {
                 await errorQueue.DeadLetter(receiveResult.PoisonMessage, connection, transaction, cancellationToken).ConfigureAwait(false);
-                return null;
             }
 
-            if (receiveResult.Successful)
+            if (receiveResult == MessageReadResult.NoMessage)
             {
-                if (await TryHandleDelayedMessage(receiveResult.Message, connection, transaction, cancellationToken).ConfigureAwait(false))
-                {
-                    return null;
-                }
-
-                return receiveResult.Message;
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Transport.SqlServer/Queuing/MessageRow.cs
```diff
@@ -45,13 +45,21 @@ public void PrepareSendCommand(SqlCommand command)
...
         {
-            return new MessageRow
+            var row = new MessageRow
             {
                 id = await dataReader.GetFieldValueAsync<Guid>(0, cancellationToken).ConfigureAwait(false),
                 expired = await dataReader.GetFieldValueAsync<int>(1, cancellationToken).ConfigureAwait(false) == 1,
                 headers = await GetHeaders(dataReader, 2, cancellationToken).ConfigureAwait(false),
                 bodyBytes = isStreamSupported ? await GetBody(dataReader, 3, cancellationToken).ConfigureAwait(false) : await GetNonStreamBody(dataReader, 5, cancellationToken).ConfigureAwait(false)
             };
+
```
SzymonPobiega, line 55: Could we move these reads to the the calling method (`ReadMessage`)? I think it would be easier to understand if this method did not advance the data reader but only read from what is there.

src/NServiceBus.Transport.SqlServer/Queuing/MessageRow.cs
```diff
@@ -45,13 +45,21 @@ public void PrepareSendCommand(SqlCommand command)
...
+            while (await dataReader.ReadAsync(cancellationToken).ConfigureAwait(false))
```
SzymonPobiega, line 56: Are these calls guaranteed to ensure that if the `DELETE` removed the row then the result of the `OUTPUT` clause has been transferred over the wire?

In other words, are we sure that they are enough to make sure that both parts of `DELETE...OUTPUT` have completed?

src/NServiceBus.Transport.SqlServer/Receiving/ProcessStrategy.cs
```diff
@@ -40,28 +40,21 @@ public void Init(TableBasedQueue inputQueue, TableBasedQueue errorQueue, OnMessa
 
         public abstract Task ProcessMessage(CancellationTokenSource stopBatchCancellationTokenSource, CancellationToken cancellationToken = default);
 
-        protected async Task<Message> TryGetMessage(SqlConnection connection, SqlTransaction transaction, CancellationTokenSource stopBatchCancellationTokenSource, CancellationToken cancellationToken = default)
+        protected async Task<MessageReadResult> TryGetMessage(SqlConnection connection, SqlTransaction transaction, CancellationTokenSource stopBatchCancellationTokenSource, CancellationToken cancellationToken = default)
```
SzymonPobiega, line 43: Since this method is much shorter now, I would actually inline it in the strategies in order to make the code easier to follow.

SzymonPobiega (changes requested):

SzymonPobiega:
Yay! It is green 🎉

SzymonPobiega (approved):

src/NServiceBus.Transport.SqlServer/Queuing/MessageRow.cs
danielmarbach: This is a bit of a nitpick but I think for the future selves it would be great to have a comment in the code explaining why this is there. Otherwise, we risk that someone thinks that is unnecessary IO and deletes the code

PR #1243, opened 2023-09-13
Prepare repo for .NET 8

andreasohlund:
Downstream follow-up to [preparing for .NET 8](https://github.com/Particular/NServiceBus/pull/6810).


andreasohlund:
@timbussmann this completes the address translation changes that you and me where part of a while back, can you take a look and see if I missed something?

andreasohlund:
@SzymonPobiega seems like transactionscopes work for .NET when running on windows

src/NServiceBus.Transport.SqlServer.TransportTests/ConfigureSqlServerTransportInfrastructure.cs
```diff
@@ -26,12 +26,10 @@ public async Task<TransportInfrastructure> Configure(TransportDefinition transpo
         inputQueueName = queueAddress.ToString();
         this.errorQueueName = errorQueueName;
 
-#if !NETFRAMEWORK
-        if (sqlServerTransport.TransportTransactionMode == TransportTransactionMode.TransactionScope)
+        if (sqlServerTransport.TransportTransactionMode == TransportTransactionMode.TransactionScope && !OperatingSystem.IsWindows())
```
andreasohlund, line 29: @bording @kentdr we could have the core transport test infrastructure do this for us?

src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -342,6 +335,10 @@ class FakePromotableResourceManager : IEnlistmentNotification
         IDelayedMessageStore delayedMessageStore = new SendOnlyDelayedMessageStore();
         TableBasedQueueCache tableBasedQueueCache;
 
+        static string DtcErrorMessage = @"
+Note that distributed transactions are not available on Linux but the transport can still used by configuring a lower transaction mode via the `SqlServerTransport.TransportTransactionMode` property when configuring the endpoint.
```
kentdr, line 339:
```suggestion
Distributed transactions are not available on Linux.  The other transaction modes can be used by setting the `SqlServerTransport.TransportTransactionMode` property when configuring the endpoint.
```
I don't know if this is better or not.  I wonder if "lower" is not the right word to use, but I can be overruled.   Also, do we typically use the markdown code block notation in error messages?

src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -342,6 +335,10 @@ class FakePromotableResourceManager : IEnlistmentNotification
...
+Be aware that different transaction modes may affect consistency guarantees as you can't rely on distributed transactions to atomically update other resources together with consuming the incoming message.";
```
kentdr, line 340:
```suggestion
Be aware that different transaction modes affect consistency guarantees since distributed transactions won't be atomically updating the resources together with consuming the incoming message.";
```

src/NServiceBus.Transport.SqlServer.AcceptanceTests/ConfigureEndpointSqlServerTransport.cs
```diff
@@ -24,9 +24,11 @@ public ConfigureEndpointSqlServerTransport()
         transport = new SqlServerTransport(connectionString);
         transport.Subscriptions.DisableCaching = true;
 
-#if !NETFRAMEWORK
-        transport.TransportTransactionMode = TransportTransactionMode.SendsAtomicWithReceive;
-#endif
+        //On Linux we need to explicitly set the transaction mode to SendsAtomicWithReceive since TransactionScope is not available there
```
timbussmann, line 27: should we be more clear that DTC is not available instead of TransactionScope?

timbussmann (approved):

src/NServiceBus.Transport.SqlServer.AcceptanceTests/ConfigureEndpointSqlServerTransport.cs
andreasohlund: agreed, updated

src/NServiceBus.Transport.SqlServer/Receiving/QueuePeeker.cs
```diff
@@ -20,7 +20,6 @@ public async Task<int> Peek(TableBasedQueue inputQueue, RepeatedFailuresOverTime
 
             try
             {
-#if NETFRAMEWORK
```
andreasohlund, line 23: @kentdr @bording I spoke to @SzymonPobiega about this and we felt that this is the most tested and safe way to ensure that connections coming out of the connection pool have the correct isolation level and since scopes now work on .net (since this does not need DTC)  we figured to use this approach which net472 has used for a long time

bording (approved):

PR #1267, opened 2023-12-14
Re-enable DTC tests and drop System.Data.SqlClient legacy transport package

bording:
With the latest preview of Microsoft.Data.SqlClient, we can now re-enable the DTC tests in this repo.

This also makes it clear that's it's time to drop the legacy transport package because it seems unlikely at this point at the bug fix will make it back into System.Data.SqlClient.


andreasohlund (approved):

danielmarbach (approved):
Yehaa. Nice! My fix works. Glad to see

PR #1279, opened 2024-01-31
Fix memory leak in translating logical addresses to SQL queue addresses

DavidBoike:
Dictionary key type `Transport.QueueAddress` does not return a deterministic hash code which results in a memory leak for items with the same key.


MarcWils:
@DavidBoike , does this only apply to v7 or are previous versions also affected? What are the reproduction steps? Does the memory leak occur when sending messages from an endpoint?

ramonsmits (approved):

DavidBoike:
@MarcWils released in 7.0.3. [Announcement here](https://discuss.particular.net/t/nservicebus-transport-sqlserver-7-0-3-patch-release-available/3827)

MarcWils:
That's fast. I'll apply the update immediatly.

PR #1305, opened 2024-03-01
Log warn when the Microsoft.Data.SqlClient version is below 5.2.0

danielmarbach:
to make sure clients start upgrading to later versions without enforcing them to have Encrypt=true that was introduced in a breaking change in 4.x of the client

### DIFF
--- a/src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
+++ b/src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
@@ -3,9 +3,15 @@ namespace NServiceBus.Transport.SqlServer
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
     using System.Transactions;
+#if SYSTEMDATASQLCLIENT
+    using System.Data.SqlClient;
+#else
+    using Microsoft.Data.SqlClient;
+#endif
     using Logging;
     using NServiceBus.Transport.SqlServer.PubSub;
     using Transport;
@@ -21,6 +27,18 @@ namespace NServiceBus.Transport.SqlServer
 
             tableBasedQueueCache = new TableBasedQueueCache(addressTranslator, !isEncrypted);
             connectionFactory = CreateConnectionFactory();
+
+            if (typeof(SqlConnection).Namespace != "Microsoft.Data.SqlClient")
+            {
+                return;
+            }
+
+            var informationalVersion = typeof(SqlConnection).Assembly.GetCustomAttributes().OfType<AssemblyInformationalVersionAttribute>().Single();
+            var currentClientVersion = new Version(informationalVersion.InformationalVersion.Split('+').First());
+            if (currentClientVersion < new Version(5, 2, 0))
+            {
+                _logger.WarnFormat("You are using an outdated version '{0}' of Microsoft.Data.SqlClient. We recommend using version 5.2.0 or later by taking an explicit package dependency to avoid issues with memory leaks or transactions running on .NET in the driver.", currentClientVersion);
+            }
         }
 
         public async Task ConfigureSubscriptions(string catalog, CancellationToken cancellationToken = default)
--- a/src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
+++ b/src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
@@ -37,7 +37,7 @@ namespace NServiceBus.Transport.SqlServer
             var currentClientVersion = new Version(informationalVersion.InformationalVersion.Split('+').First());
             if (currentClientVersion < new Version(5, 2, 0))
             {
-                _logger.WarnFormat("You are using an outdated version '{0}' of Microsoft.Data.SqlClient. We recommend using version 5.2.0 or later by taking an explicit package dependency to avoid issues with memory leaks or transactions running on .NET in the driver.", currentClientVersion);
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -21,6 +27,18 @@ internal SqlServerTransportInfrastructure(SqlServerTransport transport, HostSett
...
             tableBasedQueueCache = new TableBasedQueueCache(addressTranslator, !isEncrypted);
             connectionFactory = CreateConnectionFactory();
+
+            if (typeof(SqlConnection).Namespace != "Microsoft.Data.SqlClient")
+            {
+                return;
+            }
+
+            var informationalVersion = typeof(SqlConnection).Assembly.GetCustomAttributes().OfType<AssemblyInformationalVersionAttribute>().Single();
+            var currentClientVersion = new Version(informationalVersion.InformationalVersion.Split('+').First());
```
mauroservienti, line 37: Should these two lines be on the safer side and use `SingleOrDefault` and `FirstOrDefault`?

mauroservienti (approved):

src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
danielmarbach: In what scenario would the informational version on a microsoft shipped assembly not be present?

bording:
@danielmarbach Why 5.0.2 as a minimum specifically? Why not latest?

danielmarbach:
@bording pushed. Maybe the warning is a bit too specific now. Happy to take suggestions

src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -21,6 +27,18 @@ internal SqlServerTransportInfrastructure(SqlServerTransport transport, HostSett
...
+            if (currentClientVersion < new Version(5, 2, 0))
+            {
+                _logger.WarnFormat("You are using an outdated version '{0}' of Microsoft.Data.SqlClient. We recommend using version 5.2.0 or later by taking an explicit package dependency to avoid issues with memory leaks or transactions running on .NET in the driver.", currentClientVersion);
```
bording, line 40: Seems like we could be a little less specific here, but also point to a link that explains the connection string update that's required? That would help make it clearer as to why we aren't just upgrading the package ourselves.

Something like 
```
You are using an outdated version '{0}' of Microsoft.Data.SqlClient. We recommend using version 5.2.0 or later by adding a top-level package reference to avoid known problems in older versions of the client. See {link} for details on connection string breaking changes before ugprading.
```

danielmarbach: I'm aware of this link https://github.com/dotnet/SqlClient/blob/main/release-notes/4.0/4.0.0.md#encrypt-default-value-set-to-true or https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/7.0/microsoft-data-sqlclient-updated-to-4-0-1 but then we would be linking to 4.x version which might be confusing

danielmarbach, line 40: This might work since we link to the 5.x breaking changes then

```suggestion
                _logger.WarnFormat("You are using an outdated version '{0}' of Microsoft.Data.SqlClient. We recommend using version 5.2.0 or later by adding a top-level package reference to avoid known problems in older versions of the client. See https://github.com/dotnet/SqlClient/blob/main/release-notes/5.0/5.0.0.md for details on connection string breaking changes before ugprading.", currentClientVersion);
```

bording: Where in that link is the relevant connection string info? Seems like we should be pointing directly at that vs. some general release notes.

danielmarbach: That's the problem. It is only available here https://github.com/dotnet/SqlClient/blob/main/release-notes/4.0/4.0.0.md#encrypt-default-value-set-to-true but then again we talk about the 5.x versions in the text but link to the 4.x version which is super confusing. Then I'd rather point to the general release notes of the 5.x client or the breaking changes section within those release notes.

https://github.com/dotnet/SqlClient/blob/main/release-notes/5.0/5.0.0.md#breaking-changes

bording: > the text but link to the 4.x version which is super confusing.

I don't agree. If we need to give them information about the connection string, then a link to the 5.0 breaking changes doesn't accomplish anything.

tmasternak (approved):

src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs
```diff
@@ -21,6 +27,18 @@ internal SqlServerTransportInfrastructure(SqlServerTransport transport, HostSett
 
             tableBasedQueueCache = new TableBasedQueueCache(addressTranslator, !isEncrypted);
             connectionFactory = CreateConnectionFactory();
+
+            if (typeof(SqlConnection).Namespace != "Microsoft.Data.SqlClient")
```
poornimanayar, line 31: Is it better to use `string.Equals()` here for string comparison?

danielmarbach: Could you explain why that would be better in this case?

bording (approved):

PR #1360, opened 2024-06-12
Backport: Implement TTBR support in PostgreSQL 

tmasternak:


tmasternak:
We should not be backporting these changes to the `release-8.0`. PostgreSQL will be released as a minor.

PR #1380, opened 2024-07-04
Use docker login for ACI

danielmarbach:
This adds login for ACI to avoid throttling


danielmarbach:
FYI @DavidBoike (and yes there is redundancy and we should really have a setup action, we raised an issue about it for now)

PhilBastian (approved):

PR #1383, opened 2024-07-09
Fix bug in BackOffStrategy that can lead to infinite wait time

PhilAch:
I stumbled upon a strange behavior regarding NSB and delayed messaged on SQL transport.
Sometimes delayed messages would not be processed anymore, without any exception or anything being logged. They just stay in the .Delayed table and are not moved to their target table anymore.
If they are moved there manually, the messages are processed normally.

In the logs I found occurrences of the following error: "The value needs to be either -1 (signifying an infinite timeout), 0 or a positive integer." in BackOffStrategy.WaitForNextExecution.
This shows, that sometimes negative values can be calculated. And in case of a -1, it would lead to an infinite wait time, which could explain my noticed behavior.

Unfortunately I was not able to write a stable test to reproduce this behavior.

### DIFF
--- a/src/NServiceBus.Transport.Sql.Shared/DelayedDelivery/BackOffStrategy.cs
+++ b/src/NServiceBus.Transport.Sql.Shared/DelayedDelivery/BackOffStrategy.cs
@@ -5,7 +5,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Logging;
 
-public class BackOffStrategy(TimeProvider timeProvider = null)
+class BackOffStrategy(TimeProvider timeProvider = null)
 {
     public void RegisterNewDueTime(DateTime dueTime)
     {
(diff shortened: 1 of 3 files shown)


tmasternak:
@PhilAch thank you for your contribution. We will look into the changes and get back to you with more information.

tmasternak:
Hi @PhilAch, we took the liberty of tweaking your PR a bit i.e. adding a deterministic test, and fixing a bunch of other problems we discovered along the way.

PhilAch:
> Hi @PhilAch, we took the liberty of tweaking your PR a bit i.e. adding a deterministic test, and fixing a bunch of other problems we discovered along the way.

Hi @tmasternak, glad to see. I thought about TimeProvider, but didn't dare to change it. Anyway, happy to see it and looking forward to it being released.

tmasternak:
Hi @PhilAch, I've just realized that we are not able to run CI for externally opened PRs. As a result, I've opened https://github.com/Particular/NServiceBus.SqlServer/pull/1392 and moved all the relevant data there.

Issue #1364, opened 2024-06-13
PostgreSQL Transport


poornimanayar:
### Describe the feature.

Implementation of PostgreSQL Transport for NServiceBus.

PR #1385, opened 2024-07-16
Types from `Shared` assembly accidentally exposed as public 

tmasternak:
As part of #1364 implementation, the codebase shared between PostgreSQL and MS SQL Server transports was moved to a shared assembly. When that happened all types from that assembly should have been made internal (and visible only to PostgreSQL and MS SQL Server implementations).

This PR fixed the accessibility modifiers for accidentally exposed types to `internal`,


src/NServiceBus.Transport.Sql.Shared/Receiving/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -4,8 +4,7 @@
 using System.Threading.Tasks;
 using System.Threading;
 using System;
-
-public class RepeatedFailuresOverTimeCircuitBreaker : IDisposable
+class RepeatedFailuresOverTimeCircuitBreaker : IDisposable
```
abparticular, line 7:
```suggestion

class RepeatedFailuresOverTimeCircuitBreaker : IDisposable
```

SzymonPobiega (approved):

abparticular (approved):

Issue #1196, opened 2023-06-07
SQL Server transport requires SELECT permission when sending to remote queues


mauroservienti:
Forum thread:
- https://discuss.particular.net/t/nservicebus-8-needs-select-permission-to-send-messages/3496

### Symptoms

Endpoints sending messages to remote queues fail if the endpoint does not have `SELECT` permissions on the remote queue table. `INSERT` permissions are not enough. The endpoint send/public operations fail with an exception similar to:

```
System.Exception: Failed to send message to [TargetEndpointDB].[TargetEndpointSchema].[TargetEndpointQueue]
 ---> System.Data.SqlClient.SqlException (0x80131904): The SELECT permission was denied on the object 'TargetEndpointQueue', database 'TargetEndpointDB', schema 'TargetEndpointSchema'.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at System.Data.SqlClient.SqlDataReader.get_MetaData()
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
   at System.Data.SqlClient.SqlCommand.CompleteAsyncExecuteReader()
   at System.Data.SqlClient.SqlCommand.InternalEndExecuteReader(IAsyncResult asyncResult, String endMethod)
   at System.Data.SqlClient.SqlCommand.EndExecuteReaderInternal(IAsyncResult asyncResult)
   at System.Data.SqlClient.SqlCommand.EndExecuteReader(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
--- End of stack trace from previous location ---
[...]

PR #1440, opened 2024-10-09
Use SQL system catalog views to check for the presence of a Recoverable column

MarcWils:
Use SQL system catalog views to check for the presence of a Recoverable column. This removes the need for SELECT permissions to send a message to a queue table.

Proposal to fix:

- #1196

### DIFF
--- a/src/NServiceBus.Transport.SqlServer/Queuing/SqlServerConstants.cs
+++ b/src/NServiceBus.Transport.SqlServer/Queuing/SqlServerConstants.cs
@@ -50,7 +50,11 @@ VALUES (
 IF (@NOCOUNT = 'ON') SET NOCOUNT ON;
 IF (@NOCOUNT = 'OFF') SET NOCOUNT OFF;";
 
-        public string CheckIfTableHasRecoverableText { get; set; } = "SELECT TOP (0) * FROM {0} WITH (NOLOCK);";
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.SqlServer/Queuing/SqlServerConstants.cs
```diff
@@ -50,7 +50,11 @@ THEN DATEADD(ms, @TimeToBeReceivedMs, GETUTCDATE()) END,
 IF (@NOCOUNT = 'ON') SET NOCOUNT ON;
 IF (@NOCOUNT = 'OFF') SET NOCOUNT OFF;";
 
-        public string CheckIfTableHasRecoverableText { get; set; } = "SELECT TOP (0) * FROM {0} WITH (NOLOCK);";
+        public string CheckIfTableHasRecoverableText { get; set; } = @"
```
WilliamBZA, line 53: The challenge here is that this breaks down with multicatalog endpoints.

If an endpoint needs to send a message to a table in a separate catalog, the query needs to query the schema tables from that catalog and not the local sys.columns.

WilliamBZA:
FYI I've copied your PR to #1442 so that I can make commits to it and I've added a commit to fix the multicatalog part. We're now busy reviewing the changes to the PR. Thanks for the help!

PR #1443, opened 2024-10-10
High CPU utilization due to large number of DELETE statements

### LINKED ISSUE #1412: High CPU utilization due to large number of `DELETE` statements
### Symptoms

When using PostgreSQL transport, the database is put under a high load due to a large number of `DELETE` statements that are executed at the beginning of each received transaction.

These queries result in `0` rows being affected and are unnecessarily wasting CPU resources. Here is a sample statistics shared by one of the users affected:

image

### Who's affected

All PostgreSQL transport users.

### Root cause

The problem is caused by the peek statement overestimating the number of messages ready to be received. The query calculates the difference between "oldest" and "newest" (based on a sequence number) messages including messages that are already being received. This in turn, causes situation when a single receive transaction that takes longer than the others can cause a significant overestimation of the number of messages.

### Patched version

- 8.1.5

<details>
 <summary>Original content</summary>
#### Description
We deployed 9 net core services using NSB with Postgres which used a database  on an amazon server with 48 acu on aurora postgress serverless v2.
According to Amazon 1 ACU = 2GB RAM
https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-serverless-v2.html

Even with few data to be processed CPU did not dropped, and we were unable to keep up with processing.
I am attaching a picture with a very high number of commits and rollbacks.
image

With the database for the queues on a MSSQL Server with 8 cores and 32 GB of RAM we can run all services with less than 30% CPU.

#### Versions

  <PackageReference Include="NServiceBus" Version="9.1.1" />
  <PackageReference Include="NServiceBus.Extensions.Hosting" Version="3.0.0" />
  <PackageReference Include="NServiceBus.Persistence.Sql" Version="8.1.0" />
  <PackageReference Include="NServiceBus.Transport.PostgreSql" Version="8.1.3" />

Please list the version of the relevant packages or applications in which the bug exists.        

### Steps to reproduce

Try to see how many requests are actually done on the database.

</details>

tmasternak:
Fixes #1412


src/NServiceBus.Transport.PostgreSql.TransportTests/ConfigurePostgreSqlTransportInfrastructure.cs
```diff
@@ -11,12 +11,11 @@
 
 public class ConfigurePostgreSqlTransportInfrastructure : IConfigureTransportInfrastructure
 {
-    public TransportDefinition CreateTransportDefinition()
-    {
-        connectionString = Environment.GetEnvironmentVariable("PostgreSqlTransportConnectionString") ?? @"User ID=user;Password=admin;Host=localhost;Port=54320;Database=nservicebus;Pooling=true;Connection Lifetime=0;";
+    public static string ConnectionString =>
```
PhilBastian, line 14: is this only public so that it can be used in tests?

src/NServiceBus.Transport.PostgreSql.TransportTests/When_receive_takes_long_to_complete.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        [TestCase(TransportTransactionMode.None)]
+        [TestCase(TransportTransactionMode.ReceiveOnly)]
+        [TestCase(TransportTransactionMode.SendsAtomicWithReceive)]
+        [TestCase(TransportTransactionMode.TransactionScope)]
+        public async Task Peeker_should_provide_accurate_queue_length_estimate(TransportTransactionMode transactionMode)
+        {
+            var connectionFactory = new PostgreSqlDbConnectionFactory(ConfigurePostgreSqlTransportInfrastructure.ConnectionString);
+
+            PostgreSqlTableBasedQueue queue = await CreateATestQueue(connectionFactory);
```
PhilBastian, line 21: curious: what's your criteria for when to use var? Below you use var, even though the variable is assigned from the return of a function, similar to this like (although with a tuple)

PhilBastian (approved):

src/NServiceBus.Transport.PostgreSql.TransportTests/When_receive_takes_long_to_complete.cs
```diff
@@ -0,0 +1,95 @@
+﻿#pragma warning disable PS0018
+namespace NServiceBus.TransportTests
+{
+    using System;
+    using System.Threading;
```
PhilBastian, line 5:
```suggestion
```
marked as unnecessary

src/NServiceBus.Transport.PostgreSql.TransportTests/ConfigurePostgreSqlTransportInfrastructure.cs
tmasternak: Yes

src/NServiceBus.Transport.PostgreSql.TransportTests/When_receive_takes_long_to_complete.cs
tmasternak: Good catch. I try to use `var` whenever possible, missed that.

tmasternak:
@PhilBastian, I've had a quick review session with @andreasohlund. I think we covered all your comments. Please have a look.

andreasohlund (approved):

PR #1468, opened 2024-11-07
Stop pumps on shutdown

danielmarbach:
This aligns the transport to stop the pumps on transport shutdown and at the same time makes sure the pumps stop method can be called multiple times. 

This is part of addressing some low-hanging fruit to align all transports for the transport seam usage scenarios to make sure pumps are stopped when the transport shuts down. 

I have tried to write a transport test for it to enforce it eventually for all transport but couldn't find a way to do that without introducing flags on the transport interface and using fancy techniques like default members which seams overkill

### DIFF
--- a/src/NServiceBus.Transport.Sql.Shared/Receiving/MessageReceiver.cs
+++ b/src/NServiceBus.Transport.Sql.Shared/Receiving/MessageReceiver.cs
@@ -120,9 +120,15 @@ namespace NServiceBus.Transport.Sql.Shared.Receiving
 
         public async Task StopReceive(CancellationToken cancellationToken = default)
         {
-            messageReceivingCancellationTokenSource?.Cancel();
+            if (messageReceivingCancellationTokenSource == null)
+            {
+                // already stopped or never started
+                return;
+            }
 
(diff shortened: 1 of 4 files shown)


andreasohlund (approved):

src/NServiceBus.Transport.Sql.Shared/Receiving/MessageReceiver.cs
```diff
@@ -120,9 +120,15 @@ public async Task ChangeConcurrency(PushRuntimeSettings newLimitations,
...
         public async Task StopReceive(CancellationToken cancellationToken = default)
         {
-            messageReceivingCancellationTokenSource?.Cancel();
+            if (messageReceivingCancellationTokenSource == null)
+            {
+                // already stopped or never started
+                return;
+            }
+
+            await messageReceivingCancellationTokenSource.CancelAsync().ConfigureAwait(false);
```
tmasternak, line 129: We assume that `Stop` can be called multiple times but not concurrently. Correct?

danielmarbach: Yeah currently now message pump we have in place is concurrency safe when it comes to starting and stopping.

tmasternak (approved):

PR #1593, opened 2025-09-26
Update `When_passing_custom_connection_in_receive_context.cs`

rioka:


bording:
@rioka Thanks for the fix! I've pulled your change into #1594 so that I can get CI to run on it before I merge it to `master`.

Issue #1614, opened 2025-11-20
SQL transport runs the "Receive" query more frequently than it ought, especially when horizontally scaling


xwipeoutx:
#### Is your improvement related to a problem? Please describe.

We are using the SQL Server transport for our system, and looking at some captured SQL execution times/counts, we are noticing a lot of cases where the "Receive" command (SqlConstants.ReceiveText) is called, despite there being no messages available for processing in the queue.

This overrun is significant - we're seeing it being called 50% more times than the number of messages that end up being processes, when running in a single node.  In a multi-endpoint scenario, this is much worse, often being called N times more (where N = number of physical endpoints for the logical endpoint).

This can be seen simply with a single producer, single consumer with the following setup:

* An endpoint configured with a concurrency >= 2
* Put a message on the endpoint
* Wait for it to settle
* Count the number of times SqlConstants.ReceiveText is called - it is almost always twice, instead of once, in this scenario.

I understand the reasons here - the concurrent processing runs in the background, and it usually hasn't taken the lock out on the "receive" operation before the outer part loops around and does another "peek".

This problem is exacerbated when concurrency is higher - in my testing, it could take between 2 and 5 loops before the lock is taken out on this message.

Furthermore, this happens a lot more when scaling horizontally - obviously with multiple consumers eating through the queue, they'll query an empty queue before frequently, as the counts differ.

Ultimately, this problem prevents us from scaling horizontally while using this transport - we have headroom in our database that gets eaten up drastically in the horizontally scaled use case, and is even wasteful in the single node case.  The bottlenecks seem solvable with an improved SQL transport.

#### Describe the suggested solution

I think there could be several optimizations here that would stop the extreme DB load this causes, in both single-node and when scaling horizontally:

* For the single-node scenario, avoid processing the next peek until we're sure they're new messages.
* Prevent the next "peek" from processing until the "receive" part of the message processings have completed - or at least until some are complete (maybe configurable?)
* Track this in some way using the row_number.  I'm not sure on the impacts here with message retries, but I feel there's something to this

#### Describe alternatives you've considered

We are trying some optimisations to this transport internally - along the same vein as the above.  I understand this can impact throughput

mauroservienti:
Closing. This will be released as [8.1.11](https://github.com/Particular/NServiceBus.SqlServer/milestone/87)

PR #1619, opened 2025-11-23
Optimize peeks

xwipeoutx:
# Description

Fixes:

- #1614

This reduces the number of times the transport will run the "receive" query unnecessarily by ensuring that messages in the current "peek" batch have all at least locked the message row (via the "receive") query before doing another peek operation.

In the case of a single physical endpoint, this completely eliminates redundant reads.

In the case of multiple physical endpoints, there will still be redundant reads due to competing consumers, but this is still a significant improvement.

## Results

In testing on real-world systems, we have 300% unnecessary additional "receive" calls when the queue was empty, or all other messages were being processed. This PR makes that 0.

On a synthetic test we did, with multiple physical endpoints, each endpoint saw a similar reduction in this overrun.

## Approach

A new concept `AsyncCountdownLatch` has been added, which is similar to a `CountdownEvent` but works with async.

The message receiver creates one of these for each peek batch, with a count equal to the batch size, and passes it down to the `ProcessStrategy` being used.

The strategy will signal this latch when its receive query has been completed, and the latch will be released when all have been completed.

There is an edge case accounted for when the processing is cancelled - in this case, we want to ignore the latch, and exit the method anyway.

## Quality

All existing integration and unit tests still pass.  None extra have been written, I'm open to guidance from Particular on this.

### DIFF
--- a/src/NServiceBus.Transport.Sql.Shared/Receiving/MessageReceiver.cs
+++ b/src/NServiceBus.Transport.Sql.Shared/Receiving/MessageReceiver.cs
@@ -1,7 +1,6 @@
 namespace NServiceBus.Transport.Sql.Shared
 {
     using System;
-    using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
     using Logging;
@@ -206,7 +205,7 @@ namespace NServiceBus.Transport.Sql.Shared
                     : messageCount;
 
             bool shouldWaitForReceiveTasks = true;
-            var receiveCompletion = new CountdownEvent(maximumConcurrentProcessing);
+            var receiveLatch = new AsyncCountdownLatch(maximumConcurrentProcessing);
             for (var i = 0; i < maximumConcurrentProcessing; i++)
             {
                 if (stopBatchCancellationSource.IsCancellationRequested)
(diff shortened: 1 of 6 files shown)


src/NServiceBus.Transport.Sql.Shared/Receiving/MessageReceiver.cs
```diff
@@ -216,13 +219,19 @@ async Task ReceiveMessages(CancellationToken messageReceivingCancellationToken)
...
 
                 _ = ProcessMessagesSwallowExceptionsAndReleaseConcurrencyLimiter(stopBatchCancellationSource,
-                    localConcurrencyLimiter, messageProcessingCancellationTokenSource.Token);
+                    localConcurrencyLimiter, receiveLatch, messageProcessingCancellationTokenSource.Token);
+            }
+
+            if (shouldWaitForReceiveTasks)
+            {
+                // Wait for all receive operations to complete before returning (and thus peeking again)
+                await receiveLatch.WaitAsync().ConfigureAwait(false);
```
tmasternak: We want to pass `messageReceivingCancellationToken` here, I think.

tmasternak (commented):
Overall, this looks good to me!

One comment regarding:
> There is an edge case accounted for when the processing is cancelled - in this case, we want to ignore the latch, and exit the method anyway.

The `stopBatchCancellationSource` is used to detect a scenario when a peek saw more messages than are actually available in the queue (due to some other process reading the messages already). In such a case, we don't want to continue scheduling `receive` operations but re-execute the peek operation instead to get a better estimate on the input queue size.

Secondly, could you share which database engine (MS SQL or PostgreSQL) you are running your system on? 

Thank you :).

mauroservienti:
@xwipeoutx 

> @mauroservienti Do you want me to continue to contribute here? Or will you take the PR from here?

It'd be better if you contribute here; I'll take care of the final steps to get this merged once it's approved. We cannot run CI on this PR because it won't have access to the required secrets. So, once this is ready, I'll cherry-pick all your commits into a new PR to keep your name on them, while enabling CI to run since it'll be me triggering the build.

xwipeoutx:
Brilliant.

I'm seeing some discussion here which I can't contribute strongly too - but it seems the overall shape is fine? Are there any particulars preventing this from being approved in its current state?

src/NServiceBus.Transport.Sql.Shared/Receiving/ProcessWithNativeTransaction.cs
```diff
@@ -31,7 +32,16 @@ public override async Task ProcessMessage(CancellationTokenSource stopBatchCance
...
-                    var receiveResult = await InputQueue.TryReceive(connection, transaction, cancellationToken).ConfigureAwait(false);
+                    MessageReadResult receiveResult;
+                    try
+                    {
+                        receiveResult = await InputQueue.TryReceive(connection, transaction, cancellationToken)
+                            .ConfigureAwait(false);
+                    }
+                    finally
+                    {
+                        receiveLatch.Signal();
```
SzymonPobiega: This probably needs to be somehow moved up because in the current shape if an exception happens when opening connection or transaction and all other concurrent receives succeed, the counter will be missing one element and won't let the peek continue.

PR #1640, opened 2025-12-10
Use a disposable object to encapsulate the logic of signaling the latch when an exception is thrown in the processing path

SzymonPobiega:
Use a disposable object to encapsulate the logic of signaling the latch when an exception is thrown in the processing path


src/NServiceBus.Transport.Sql.Shared/Receiving/MessageReceiver.cs
mauroservienti:
```suggestion
```

PR #1725, opened 2026-06-09
Refactor SHA256 hash computation for better resource management

danielmarbach:
This pull request refactors the `QueueCreator` class in the `NServiceBus.Transport.PostgreSql` namespace to use C# primary constructor syntax and makes a minor update to the hash calculation method. The changes also shift the logger to a static readonly field and clean up field declarations.

Class refactoring and initialization:

* Refactored `QueueCreator` to use a primary constructor, removing explicit field declarations and the constructor body. This simplifies the code and leverages modern C# syntax.

Logging:

* Changed the logger declaration to a static readonly field, moving it to the bottom of the class for clarity and to match conventions.

Hash calculation update:

* Updated the SHA-256 hash calculation in `CalculateLockId` to use `SHA256.HashData` instead of creating a new SHA256 instance, improving efficiency and clarity.

### DIFF
--- a/src/NServiceBus.Transport.PostgreSql/Receiving/QueueCreator.cs
+++ b/src/NServiceBus.Transport.PostgreSql/Receiving/QueueCreator.cs
@@ -91,7 +91,8 @@ namespace NServiceBus.Transport.PostgreSql
         static long CalculateLockId(string text)
         {
             var byteContents = Encoding.Unicode.GetBytes(text);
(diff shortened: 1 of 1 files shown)


danielmarbach:
This should probably never have used SHA256 but given it is not on a high-performance path, I'm leaving this out of scope for now

SzymonPobiega (approved):

ramonsmits (approved):

PR #1731, opened 2026-06-09
Concurrent delayed delivery backoff scheduling race

danielmarbach:
Fixes a race in `BackOffStrategy` where delayed-message due-time notifications could concurrently mutate shared scheduling state while the delayed delivery processor was calculating or waiting for the next execution time.

The state is now updated atomically, so `NextExecutionTime`, `NextDelayedMessage`, delayed-message availability, and the exponential backoff interval remain consistent when `RegisterNewDueTime` is called concurrently with WaitForNextExecution.

Added coverage for:

- concurrent due-time registrations preserving the earliest execution time
- due-time registration while the processor is already waiting
- existing backoff behavior using FakeTimeProvider instead of wall-clock delays

Also converted the existing `BackOffStrategyTests` wait-based cases to advance fake time deterministically, reducing the focused test runtime from seconds to milliseconds.

## Before

image

## After

image

### DIFF
--- a/src/NServiceBus.Transport.Sql.Shared/DelayedDelivery/BackOffStrategy.cs
+++ b/src/NServiceBus.Transport.Sql.Shared/DelayedDelivery/BackOffStrategy.cs
@@ -5,101 +5,147 @@ using System.Threading;
 using System.Threading.Tasks;
 using Logging;
 
-class BackOffStrategy
+class BackOffStrategy(TimeProvider timeProvider)
 {
     public BackOffStrategy() : this(TimeProvider.System) { }
 
-    public BackOffStrategy(TimeProvider timeProvider)
-    {
-        this.timeProvider = timeProvider;
-
-        NextExecutionTime = timeProvider.GetUtcNow().UtcDateTime.AddSeconds(2);
-        NextDelayedMessage = timeProvider.GetUtcNow().UtcDateTime;
-    }
-
-    public void RegisterNewDueTime(DateTime dueTime)
-    {
-        NextDelayedMessage = dueTime;
-
-        if (dueTime == DateTime.MinValue)
+    public void RegisterNewDueTime(DateTime dueTime) =>
+        UpdateState(static (current, dueTime) =>
         {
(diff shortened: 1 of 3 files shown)


SzymonPobiega:
@danielmarbach do you need my review here or am I optional? I am asking because it's been years since I last looked at this class and a reliable review would require me to pull the sources and understand the new version. Comparing the diffs is really not useful here.

ramonsmits (approved):
The magic happens in `UpdateState` and uses `CompareExchange` instead of `lock` which is fine as this is likely low contention. Either would be an improvement over the previous version.

danielmarbach:
> @danielmarbach do you need my review here or am I optional? I am asking because it's been years since I last looked at this class and a reliable review would require me to pull the sources and understand the new version. Comparing the diffs is really not useful here.

@SzymonPobiega no worries

tmasternak (approved):
I wonder if we should consider moving the NextExecutionTime calculation and management to the ReceiveStrategy, which is the single user of the BackOffStrategy. This would make the state simpler (only NextDelayedMessage), which could make it simpler to understand.

danielmarbach:
> I wonder if we should consider moving the NextExecutionTime calculation and management to the ReceiveStrategy, which is the single user of the BackOffStrategy. This would make the state simpler (only NextDelayedMessage), which could make it simpler to understand.

I looked at this briefly and it seems it would shift around some complexity while the back of strategy would get simpler, the receive strategy becomes more complex, and it's harder to test compared to the backoff strategy, but your input gave me an idea which I'm trying to look at, which should actually simplify the backoff strategy quite a bit.

danielmarbach:
Scratch that. I thought a few things through and I don't think this actually simplifies things. The idea was essentially using two longs to represent the DateTimes, but you'd still need volatile reads and compare-exchange loops — so the complexity doesn't go away, it just moves around. On top of that, splitting the state across two longs introduces race windows that the current implementation simply doesn't have. The existing code nicely encapsulates everything into domain objects and helper utilities that I think make the intent clear and keep the correctness guarantees intact.

PR #1726, opened 2026-06-09
Refactor DictionarySerializer for improved memory stream handling

danielmarbach:
This pull request refactors the `DictionarySerializer` class to modernize its code style and improve performance in serialization and deserialization. The main changes include adopting C# 8.0+ syntax for `using` statements, optimizing buffer usage, and ensuring immutability for the serializer instance.

**Code style and performance improvements:**

* Replaced the traditional `using` statement block with the newer C# syntax (`using var`) for `MemoryStream` in both `Serialize` and `DeSerialize` methods, making the code more concise and readable.
* Optimized the serialization process by using `TryGetBuffer` to avoid unnecessary allocations when converting the stream to a UTF-8 string.
* Made the `serializer` field `readonly` and used target-typed `new` for cleaner and safer initialization.

PR #1733, opened 2026-06-10
Modernize DictionarySerializer to use System.Text.Json in a backward compatible way

danielmarbach:
This PR follows up on #1726 by replacing DataContractJsonSerializer with source-generated System.Text.Json for header serialization.

The primary goal of this change is to modernize the implementation while preserving compatibility with existing persisted data. Additional tests were added to capture the exact behavior of the previous serializer, including edge cases around escaping and Unicode handling. The new implementation was then adjusted to produce the same serialized output and deserialization behavior.

Benefits of this change include:
- Native AOT and trimming compatibility through source generation
- Reduced reliance on legacy serialization APIs
- Improved performance characteristics of System.Text.Json
- Preserved compatibility with headers serialized by previous versions of the transport

The commits in this PR intentionally separate the work:
1. Capture and document the existing DataContractJsonSerializer behavior through tests and approvals.
2. Introduce the System.Text.Json implementation.
3. Adjust the implementation to maintain backward compatibility with the previously persisted JSON representation.


danielmarbach:
The first commit runs the tests against the data contract serializer and writes that into approved files. Then I changed the serializer and iterated on it until it was 100% compatible with the tests.

DavidBoike (approved):

src/NServiceBus.Transport.Sql.Shared/Queuing/DictionarySerializer.cs
```diff
@@ -1,28 +1,49 @@
...
+                foreach (var c in source)
+                {
+                    if (c == '/')
+                    {
+                        destination[position++] = '\\';
+                    }
+
+                    destination[position++] = c;
+                }
+            });
```
ramonsmits: Instead of a per char copy... use CopyTo which is SIMD friendly.

Simply iterate over the chunks and CopyTo per chunk. If this is faster depends on the XML but slashed are rare to I think on average it will be.



```suggestion
						return string.Create(json.Length + slashCount, json, static (destination, source) =>
						    {
						        var remaining = source.AsSpan();
						        int idx;
						        while ((idx = remaining.IndexOf('/')) >= 0)
						        {
						            remaining[..idx].CopyTo(destination);
						            destination[idx] = '\\';
						            destination[idx + 1] = '/';
						            destination = destination[(idx + 2)..];
						            remaining = remaining[(idx + 1)..];
						        }
						        remaining.CopyTo(destination);
						    });
```

src/NServiceBus.Transport.Sql.Shared/Queuing/DictionarySerializer.cs
```diff
@@ -1,28 +1,49 @@
...
-        public static Dictionary<string, string> DeSerialize(string json)
+        public static Dictionary<string, string>? Deserialize(string value)
+            => JsonSerializer.Deserialize(value, Context.DictionaryStringString);
+
+        static string EscapeDataContractCompatible(string json)
         {
-            using var stream = new MemoryStream(Encoding.UTF8.GetBytes(json));
-            return (Dictionary<string, string>)serializer.ReadObject(stream);
+            var span = json.AsSpan();
+            var slashCount = span.Count('/');
```
ramonsmits, line 22: @danielmarbach at first I thought why don't you use IndexOf until I saw the count used later :-)

ramonsmits (dismissed):
string create routing can be optimized a little more.

tmasternak (approved):

PR #1756, opened 2026-07-11
Record transport transaction state explicitly instead of inferring it

dvdstelt:
## Why

I always had issues reading the many if-statements on what state/context we're in. So I created this and hope it's more readable from now on.

## What

`MessageDispatcher.DispatchDefault` reverse-engineered the transaction mode by probing which entries happened to be present in the `TransportTransaction` bag, through an order-dependent if/else chain of extension methods (`OutsideOfHandler`, `IsNoTransaction`, `IsReceiveOnly`, `IsSendsAtomicWithReceive`, `IsTransactionScope`). Every `TransportTransaction` is created by code that already knows the mode, so this PR stamps a `TransportTransactionState` into the transaction at creation time and turns the dispatcher into a single exhaustive switch.

- New internal `TransportTransactionState` enum; all `TransportTransactions` factory methods record it.
- `DispatchDefault` switches on the state, with a comment per case explaining how sends relate to the receive transaction. The two identical "open a dedicated connection" blocks (outside handler and receive-only) collapse into one case.
- Inference from bag contents remains only as a fallback (`InferState`) for instances not created by the transport: the empty one core creates for dispatches outside the message processing pipeline, and hand-rolled ones used by external integrations. It mirrors the old chain's semantics, including its ordering.
- The `"SqlTransport.ReceiveOnlyTransactionMode"` key was defined in two classes; the writer used one copy and the readers the other, working only because the strings happened to match. It now lives once in `TransportTransactionKeys` and is still written to the bag for downstream components (e.g. SQL persistence).
- `SendOptionsExtensions` hand-assembled user-provided transactions with its own copy of the key strings while `PublishOptionsExtensions` used the `TransportTransactions.UserProvided(...)` factory; both now use the factory, and the duplicate key was removed from `SettingsKeys`.

## Behavior notes

No public API changes (API approval tests unchanged). One deliberate alignment: a user-provided connection-only transaction reaching default (non-immediate) dispatch previously got wrapped in a new transaction, while the same bag in isolated dispatch did not. That path is unreachable in practice because `UseCustomSqlConnection`/`UseCustomSqlTransaction` force immediate dispatch; both paths now behave identically (dispatch on the user's connection). Everything else is a faithful translation of the old chain.


dvdstelt:
ping

danielmarbach:
I pushed what I believe are simplifications and one commit that contains a proposal that would switch to record types that in the future we could then use with C sharp discriminated union support in C sharp 15 we could create a closed hierarchy. It has one downside, it creates one class allocation more compared to the state approach. But given the complexity of this connection handling and that it makes it more expressive, it might be worthwhile the trade-off. But that's why I called the commit discussion for the proposal, and if we decide, we can just drop it. Do you have any thoughts, @dvdstelt ?

dvdstelt:
I like the records, mostly for the non-null connection/transaction per case rather than the future C# 15 exhaustiveness. But storing the state next to SqlConnection/SqlTransaction reintroduced the two-sources-of-truth problem your previous commit removed.

So I pushed two commits on top:

- State is now derived on read instead of stored, so the entries stay the single source of truth. MessageDispatcher needed no changes.
- `OutsideHandler`, `ReceiveOnly` and `AmbientTransaction` became empty singletons since the dispatcher opens its own connection for those anyway. Only 3 of the 6 states allocate now, so the allocation downside mostly disappears.
- Added unit tests for the state derivation, which we only covered through integration tests before.

If you're OK with this shape, let's merge. Although there's still no review.

danielmarbach (approved):

